PR #3, opened 2022-02-08
Introducing OpenAPI Generator

ricardozanini:
Minimal use case to add OpenAPI Rest Client generation for Quarkus.


deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/GeneratedOpenApiFile.java
```diff
@@ -0,0 +1,17 @@
+package io.quarkiverse.openapi.generator.deployment;
+
+import org.jboss.jandex.ClassInfo;
+
+import io.quarkus.builder.item.MultiBuildItem;
+
+public class GeneratedOpenApiFile extends MultiBuildItem {
```
gastaldi, line 7: Classes extending BuildItems should always be final. If this is meant to be reused, make it abstract

deployment/pom.xml
```diff
@@ -9,16 +9,137 @@
...
+          <groupId>javax.validation</groupId>
+          <artifactId>validation-api</artifactId>
+        </exclusion>
+      </exclusions>
+    </dependency>
+    <dependency>
+      <groupId>org.slf4j</groupId>
+      <artifactId>slf4j-ext</artifactId>
+      <version>${version.org.slf4j}</version>
+    </dependency>
```
gastaldi, line 134: Do you really need SL4J here? The best is to use the JBoss Logging adapters instead

deployment/pom.xml
```diff
@@ -9,16 +9,137 @@
...
+    <dependency>
+      <groupId>io.quarkus</groupId>
+      <artifactId>quarkus-rest-client-jackson-deployment</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>io.quarkus</groupId>
+      <artifactId>quarkus-devtools-utilities</artifactId>
+      <!-- Maybe we need to add it to the quarkus bom -->
+      <version>${quarkus.version}</version>
```
gastaldi, line 35: https://github.com/quarkusio/quarkus/commit/7526f31d3386517f351fff7616e841f406acec5e introduced it in Quarkus 2.7.1.Final

deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/codegen/OpenApiGeneratorCodeGenBase.java
```diff
@@ -0,0 +1,83 @@
...
+        }
+        return false;
+    }
+
+    private String escapeWhitespace(String path) {
+        if (OS.determineOS() == OS.LINUX) {
+            return path.replace(" ", "\\ ");
+        } else {
+            return path;
+        }
```
gastaldi, line 72: Don't you also need to escape on Windows?

deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/template/QuteTemplatingEngineAdapter.java
```diff
@@ -0,0 +1,73 @@
...
+    };
+    public final Engine engine;
+
+    public QuteTemplatingEngineAdapter() {
+        this.engine = Engine.builder()
+                .addDefaults()
+                .addValueResolver(new ReflectionValueResolver())
+                .removeStandaloneLines(true)
+                .strictRendering(false)
+                .build();
```
gastaldi, line 39: When running in Quarkus, the Engine is available as a CDI managed object. Maybe lookup first before attempting to create a new one?

gastaldi (commented):
Added some comments. Looks good overall

deployment/pom.xml
ricardozanini: Unfortunately, it's used internally by the underlying openapi library: https://github.com/OpenAPITools/openapi-generator/blob/master/modules/openapi-generator/src/main/java/org/openapitools/codegen/utils/OnceLogger.java#L10. 

What I'm doing here is aligning with the Quarkus versions, at least...

deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/codegen/OpenApiGeneratorCodeGenBase.java
ricardozanini: This code has been borrowed from grpc, I suppose it's to cover a corner case on Linux: https://github.com/quarkusio/quarkus/blob/main/extensions/grpc/codegen/src/main/java/io/quarkus/grpc/deployment/GrpcCodeGen.java#L252-L258

@michalszynkiewicz can you clarify? :)

deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/template/QuteTemplatingEngineAdapter.java
ricardozanini: The CDI engine is unavailable at this stage (CodeGenProvider): I suppose it's running even before the processor. So that's why I'm manually creating it.

integration-tests/src/test/java/io/quarkiverse/openapi/generator/it/PetStoreTest.java
```diff
@@ -0,0 +1,24 @@
...
+import io.quarkus.test.common.QuarkusTestResource;
+import io.quarkus.test.junit.QuarkusTest;
+
+@QuarkusTestResource(WiremockPetStore.class)
+@QuarkusTest
+public class PetStoreTest {
+
+    @Test
+    public void testGetPetById() {
+        final String petName = when()
```
gastaldi, line 17: Not sure what you're testing here. Shouldn't the REST Client be injected instead of using rest-assured?

ricardozanini: It was a copy and paste from the later test inside the `deployment` module. I'll change.

gastaldi:
Does it integrate with the Quarkus OpenAPI extension? Eg. Can you generate a client based on the output of `http://localhost:8089/openapi`?

ricardozanini:
It does not integrate automatically but can be a feature in the roadmap. :)

If what you're asking is to use as an input a remote reference like `https://petstore.swagger.io/v2/swagger.json`, it can't be done now, but I'm willing to implement since my project will use it. See: https://github.com/serverlessworkflow/specification/blob/0.8.x/specification.md#using-functions-for-restful-service-invocations

deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/OpenApiGeneratorProcessor.java
```diff
@@ -0,0 +1,39 @@
...
+            CombinedIndexBuildItem index) {
+
+        boolean hasGeneratedFiles = false;
+        for (SpecConfig spec : configuration.specs.values()) {
+            for (ClassInfo classInfo : index.getIndex().getKnownClasses()) {
+                if (classInfo.name().packagePrefix().equals(spec.apiPackage)) {
+                    restClients.produce(new GeneratedOpenApiRestClientBuildItem(classInfo));
+                    hasGeneratedFiles = true;
+                } else if (classInfo.name().packagePrefix().equals(spec.modelPackage)) {
+                    models.produce(new GeneratedOpenApiModelBuildItem(classInfo));
```
michalszynkiewicz, line 29: maybe I'm misreading it, but wouldn't it make slightly more sense to have only one package name defined and created a name for the other by appending `.model`, or just use one package for both types of generated things?
It would simplify the config for users

ricardozanini: I was trying to mimic the exact configuration of the underlying library, but I agree that having only one property would simplify things for users. Let's do it.

deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/OpenApiGeneratorConfiguration.java
```diff
@@ -0,0 +1,28 @@
...
+public class OpenApiGeneratorConfiguration {
+    /**
+     * Fine tune the configuration for each OpenAPI spec file in `src/openapi` directory.
+     * <p>
+     * The file name is used to index this configuration. For example:
+     * `quarkus.openapi-generator.spec."myfilespec.json".api-package=org.acme.api`.
+     * <p>
+     * If you have more than one file to generate rest clients, it is **highly** recommended that you add this configuration
+     * to your properties file to make a distinction between the generated APIs.
+     **/
```
michalszynkiewicz, line 19: for me this means the config is not required but if I read the generator properly, it actually requires this setting

deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/codegen/OpenApiGeneratorCodeGenBase.java
michalszynkiewicz: that's what protoc needed, it was specific to linux. I think we should try with openapi generator first before adding escaping

ricardozanini: I'll run some tests with a file with a space to see how that goes.

gastaldi: You might want to change the workflow action to build on Windows too

integration-tests/src/main/resources/application.properties
```diff
@@ -1,2 +1,3 @@
 quarkus.openapi-generator.spec."petstore.json".base-package=org.acme.openapi
-quarkus.openapi-generator.spec."openweather.yaml".base-package=org.acme.openapi.weather
\ No newline at end of file
+# since the file name has a space, we use the URI representation of this space here to not break the properties file
+quarkus.openapi-generator.spec."open%20weather.yaml".base-package=org.acme.openapi.weather
```
michalszynkiewicz, line 3: maybe we could allow "open weather.yaml" and encode for the user it as it's done for paths?

michalszynkiewicz (approved):

integration-tests/src/main/resources/application.properties
ricardozanini: @michalszynkiewicz I had problems adding properties with spaces. It won't get read correctly. That's why I kept the encoding here to show how end-users should do this.

ricardozanini: btw, we have a nice wrong message for them as well:

```
The config property quarkus.openapi-generator.spec."open%20weather.yaml".base-package is required but it could not be found in any config source
```

ricardozanini:
> 

More or less the same thing, yes. I want to release a first version of the extension supporting a simple use case (sync REST) and adding more as we go and receive feedback.

PR #20, opened 2022-02-25
Added Circuit Breaker feature

hbelmiro:
Please see the new Circuit Breaker section added to README.md to know how to use the new feature.


README.md
```diff
@@ -96,6 +96,114 @@ public class PetResource {
...
+quarkus.openapi-generator.CircuitBreaker.enabled=true
+
+# Defines the configuration for the GET method to the /bye endpoint
+quarkus.openapi-generator.spec."simple-openapi.json"/byeGet/CircuitBreaker/failOn = java.lang.IllegalArgumentException,java.lang.NullPointerException
+quarkus.openapi-generator.spec."simple-openapi.json"/byeGet/CircuitBreaker/skipOn = java.lang.NumberFormatException
+quarkus.openapi-generator.spec."simple-openapi.json"/byeGet/CircuitBreaker/delay = 33
+quarkus.openapi-generator.spec."simple-openapi.json"/byeGet/CircuitBreaker/delayUnit = MILLIS
+quarkus.openapi-generator.spec."simple-openapi.json"/byeGet/CircuitBreaker/requestVolumeThreshold = 42
+quarkus.openapi-generator.spec."simple-openapi.json"/byeGet/CircuitBreaker/failureRatio = 3.14
+quarkus.openapi-generator.spec."simple-openapi.json"/byeGet/CircuitBreaker/successThreshold = 22
```
ricardozanini, line 176: Hi! Instead of replicating the properties from the Circuit Breaker extension, what do you think about reusing them?

See: https://quarkus.io/guides/smallrye-fault-tolerance#runtime-configuration

In this case, the properties would be:

```properties
org.acme.openapi.simple.DefaultApi/byeGet/CircuitBreaker/failOn = java.lang.IllegalArgumentException,java.lang.NullPointerException
org.acme.openapi.simple.DefaultApi/byeGet/CircuitBreaker/skipOn = java.lang.NumberFormatException
org.acme.openapi.simple.DefaultApi/byeGet/CircuitBreaker/delay = 33
org.acme.openapi.simple.DefaultApi/byeGet/CircuitBreaker/delayUnit = MILLIS
org.acme.openapi.simple.DefaultApi/byeGet/CircuitBreaker/requestVolumeThreshold = 42
org.acme.openapi.simple.DefaultApi/byeGet/CircuitBreaker/failureRatio = 3.14
org.acme.openapi.simple.DefaultApi/byeGet/CircuitBreaker/successThreshold = 22
```

This way we delegate to the proper extension to do the right configuration.

The generated code then will have only the annotations for `CircuitBreaker`s, `Retry`, etc:

```java
@Path("")
@RegisterRestClient
public interface DefaultApi {
    @GET
    @Path("/bye")
    @Produces({"text/plain"})
    @CircuitBreaker() //simple annotation, the properties are fine tuned via properties file
    public String byeGet();
    @GET
    @Path("/hello")
    @Produces({"text/plain"})
    public String helloGet();
}
```

We just need to explain how the class fqdn is generated (`base-package` + tag + Api prefix), so the user would know how to set the properties. Alternatively, they can generate the code first and add the properties later having the final class generated. :)

ricardozanini (commented):
Many thanks for this PR!! I really appreciated it. I have one suggestion, though. Please see the comment above.

README.md
hbelmiro: Actually, it was my initial idea and I thought the user having to know the FQCN would be a problem.
But on second thought, it is a good price to pay to delegate almost everything to SmallRye.
I switched the PR to draft until I do this change.

hbelmiro:
@ricardozanini I committed the [suggested changes](https://github.com/quarkiverse/quarkus-openapi-generator/pull/20#discussion_r815130537)

hbelmiro:
@fjtirado can you also take a look at this, please?

deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/circuitbreaker/CircuitBreakerConfigurationParser.java
```diff
@@ -0,0 +1,59 @@
...
+import java.util.function.UnaryOperator;
+import java.util.stream.Collectors;
+
+import io.quarkiverse.openapi.generator.deployment.circuitbreaker.CircuitBreakerConfiguration.CircuitBreakerClassConfiguration;
+
+public final class CircuitBreakerConfigurationParser {
+
+    private final UnaryOperator<String> nameToValuePropertyMapper;
+
+    public CircuitBreakerConfigurationParser(UnaryOperator<String> nameToValuePropertyMapper) {
```
fjtirado, line 15: I think it will be better to have a constructor that accepts both the UnaryOperator and the Collection of propery names.

deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/codegen/OpenApiGeneratorCodeGenBase.java
```diff
@@ -61,8 +73,21 @@ public boolean trigger(CodeGenContext context) throws CodeGenException {
         return false;
     }
 
+    private CircuitBreakerConfiguration getCircuitBreakerConfiguration(CodeGenContext context) {
+        UnaryOperator<String> nameToValuePropertyMapper = propertyName -> context.config().getValue(propertyName,
```
fjtirado, line 77: If you have the static method with config  arguments, you can write
`return CircuitBreakerConfigurationParser.of(context.config());`

deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/codegen/OpenApiGeneratorCodeGenBase.java
```diff
@@ -61,8 +73,21 @@ public boolean trigger(CodeGenContext context) throws CodeGenException {
...
+                .parse(getConfigPropertyNames(context));
+    }
+
     private String getRequiredIndexedProperty(final String propertyKey, final CodeGenContext context) {
         // this is how we get a required property. The configSource will handle the exception for us.
         return context.config().getValue(propertyKey, String.class);
     }
+
+    private static List<String> getConfigPropertyNames(CodeGenContext context) {
+        return StreamSupport.stream(context.config().getPropertyNames().spliterator(), false)
```
fjtirado, line 90: This method can be removed if other changes are done

deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/circuitbreaker/CircuitBreakerConfigurationParser.java
```diff
@@ -0,0 +1,59 @@
...
+        this.nameToValuePropertyMapper = nameToValuePropertyMapper;
```
fjtirado, line 16: I think is better to have a static method that just acceps the `Config` object

`public static CircuitBreakerConfiguration of (Config config) {
return new CircuirBreakConfiguration(readClasses(config));
}
`

deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/circuitbreaker/CircuitBreakerConfigurationParser.java
```diff
@@ -0,0 +1,59 @@
...
+import java.util.List;
+import java.util.Set;
+import java.util.function.UnaryOperator;
+import java.util.stream.Collectors;
+
+import io.quarkiverse.openapi.generator.deployment.circuitbreaker.CircuitBreakerConfiguration.CircuitBreakerClassConfiguration;
+
+public final class CircuitBreakerConfigurationParser {
+
+    private final UnaryOperator<String> nameToValuePropertyMapper;
```
fjtirado, line 13: This can be removed

deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/circuitbreaker/CircuitBreakerConfigurationParser.java
```diff
@@ -0,0 +1,59 @@
...
+    private List<CircuitBreakerClassConfiguration> readClasses(Collection<String> propertyNames) {
+        List<String> filteredPropertyNames = filterPropertyNames(propertyNames).stream()
+                .filter(property -> nameToValuePropertyMapper.apply(property).equals("true"))
+                .collect(Collectors.toList());
+
+        Set<String> classNames = filteredPropertyNames.stream()
+                .map(this::getClassName)
+                .collect(Collectors.toSet());
+
+        return classNames.stream()
```
fjtirado, line 32: When using config, You can rewrite the whole method this way (still using functional approach)
```
 List<String> filteredPropertyNames = StreamSupport.stream(config.getPropertyNames().spliterator()).filter(property->config.getConfigValue(property,Boolean.class).collect(Collectors.toList());

return filteredPropertyNames.stream().map(this::getClassName).collect(Collectors.toMap(c ->c, c->getMethodNames(c,filteredPropertyNames));
```

deployment/src/test/java/io/quarkiverse/openapi/generator/deployment/assertions/Assertions.java
```diff
@@ -0,0 +1,15 @@
...
+import com.github.javaparser.ast.body.MethodDeclaration;
+
+import io.quarkiverse.openapi.generator.testutils.circuitbreaker.assertions.CircuitBreakerMethodAssert;
+
+public final class Assertions extends org.assertj.core.api.Assertions {
+
+    private Assertions() {
+    }
+
+    public static CircuitBreakerMethodAssert assertThat(MethodDeclaration actual) {
```
fjtirado, line 12: Do you need that class? Why not call directly the references classs?

integration-tests/src/test/java/io/quarkiverse/openapi/generator/it/assertions/Assertions.java
```diff
@@ -0,0 +1,15 @@
...
+import com.github.javaparser.ast.body.MethodDeclaration;
+
+import io.quarkiverse.openapi.generator.testutils.circuitbreaker.assertions.CircuitBreakerMethodAssert;
+
+public final class Assertions extends org.assertj.core.api.Assertions {
+
+    private Assertions() {
+    }
+
+    public static CircuitBreakerMethodAssert assertThat(MethodDeclaration actual) {
```
fjtirado, line 12: Same, do we really need that class?

pom.xml
```diff
@@ -31,6 +36,24 @@
...
+      <dependency>
+        <groupId>com.github.javaparser</groupId>
+        <artifactId>javaparser-core</artifactId>
+        <version>${version.com.github.javaparser}</version>
+      </dependency>
+
+      <dependency>
+        <groupId>org.assertj</groupId>
+        <artifactId>assertj-core</artifactId>
+        <version>${version.org.assertj}</version>
```
fjtirado, line 49: This should be scope test, isnt it?

deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/circuitbreaker/CircuitBreakerConfigurationParser.java
```diff
@@ -0,0 +1,59 @@
...
+
+    public CircuitBreakerConfigurationParser(UnaryOperator<String> nameToValuePropertyMapper) {
+        this.nameToValuePropertyMapper = nameToValuePropertyMapper;
+    }
+
+    public CircuitBreakerConfiguration parse(Collection<String> propertyNames) {
+        return new CircuitBreakerConfiguration(readClasses(propertyNames));
+    }
+
+    private List<CircuitBreakerClassConfiguration> readClasses(Collection<String> propertyNames) {
```
fjtirado, line 23: change signature to
`private static  Map<String,List<String>> readClasses(Config config) {`

deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/circuitbreaker/CircuitBreakerConfiguration.java
```diff
@@ -0,0 +1,43 @@
...
+
+    public CircuitBreakerConfiguration(List<CircuitBreakerClassConfiguration> classConfigurations) {
+        this.classConfigurations = List.copyOf(classConfigurations);
+    }
+
+    public List<CircuitBreakerClassConfiguration> getClassConfigurations() {
+        return classConfigurations;
+    }
+
+    CircuitBreakerClassConfiguration getClassConfiguration(String className) {
```
fjtirado, line 17: I think is better to make this method public

deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/circuitbreaker/CircuitBreakerConfiguration.java
```diff
@@ -0,0 +1,43 @@
+package io.quarkiverse.openapi.generator.deployment.circuitbreaker;
+
+import java.util.List;
+
+public final class CircuitBreakerConfiguration {
+
+    private final List<CircuitBreakerClassConfiguration> classConfigurations;
```
fjtirado, line 7: I think this can be replaced by 
private final Map<String, List<String> classConfigurations;

deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/circuitbreaker/CircuitBreakerConfiguration.java
fjtirado: The implemenation with map will be classConfiguration.get(className), contant rather than lineal

deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/circuitbreaker/CircuitBreakerConfiguration.java
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    CircuitBreakerClassConfiguration getClassConfiguration(String className) {
+        return getClassConfigurations().stream()
+                .filter(c -> c.getClassName().equals(className))
+                .findFirst()
+                .orElseThrow(() -> new IllegalArgumentException("Configuration not found for class: " + className));
+    }
+
+    public static final class CircuitBreakerClassConfiguration {
```
fjtirado, line 24: This class can be removed if using a map

deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/circuitbreaker/CircuitBreakerConfiguration.java
```diff
@@ -0,0 +1,43 @@
+package io.quarkiverse.openapi.generator.deployment.circuitbreaker;
+
+import java.util.List;
+
+public final class CircuitBreakerConfiguration {
+
+    private final List<CircuitBreakerClassConfiguration> classConfigurations;
+
+    public CircuitBreakerConfiguration(List<CircuitBreakerClassConfiguration> classConfigurations) {
```
fjtirado, line 9: If using a map, the constructor should be changed accordingly. 
In any case, you do need to call copyOf

fjtirado (commented):
@hbelmiro Please take a look at the proposed changes
I think the config can be expressed a map where classname is the key and the list of method names is the value.

deployment/src/main/resources/templates/api.qute
```diff
@@ -53,6 +53,11 @@ public interface {classname} {
     {#if op.hasProduces}
     @Produces(\{{#for produce in op.produces}"{produce.mediaType}"{#if produce_hasNext}, {/if}{/for}\})
     {/if}
+    {#for cbClassConfig in circuit-breaker.classConfigurations.orEmpty}{#if cbClassConfig.className == package + classname}
```
fjtirado, line 56: If using a map, I think this will still work, you just need to change
cbClassConfig.className by cbClassConfig.key
and
cbClassConfig.methods by cbClassConfig.value

deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/circuitbreaker/CircuitBreakerConfiguration.java
fjtirado: If you feel that the configuration might contain more thing than a list of method names in the future. Then keep the map (becuase classname is the key to quick access in any case) and change the value to a POJO rather than a list. 
If there is not need for a map at all, because the configuration is processed always as a loop, then the package scoped method should be removed, but I think you need a map, my only doubt is if the value will be a pojo (containing the list of method names and more info) rather than just the list. I vote for using just list now.

deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/circuitbreaker/CircuitBreakerConfigurationParser.java
hbelmiro: I agree it would make the code much simpler, but it would also couple this class with MicroProfile Config. Besides that, it would make it harder to test, since it would need a Config.
WDYT?

deployment/src/test/java/io/quarkiverse/openapi/generator/deployment/assertions/Assertions.java
hbelmiro: Without that class, we would have to qualify the calls to `CircuitBreakerMethodAssert#assertThat`, since we wouldn't be able to statically import `assertThat` from `CircuitBreakerMethodAssert` and from `org.assertj.core.api.Assertions` at the same type.

Without that class the code would be like the following:

````java
import static org.assertj.core.api.Assertions.assertThat;
import io.quarkiverse.openapi.generator.testutils.circuitbreaker.assertions.CircuitBreakerMethodAssert;

@Test
void test() {
        // ...   

        // pure AssertJ assertion
        assertThat(byeMethod).isNotEmpty();

        // custom assertion
        CircuitBreakerMethodAssert.assertThat(byeMethod.orElseThrow())
                .hasCircuitBreakerAnnotation()
                .doesNotHaveAnyCircuitBreakerAttribute();

        // ...   
}
````

With that class the code would be like the following:

````java
// Our custom class includes the pure AsserJ assertions
import static io.quarkiverse.openapi.generator.deployment.assertions.Assertions.assertThat;

@Test
void test() {
        // ...   

        // pure AssertJ assertion
        assertThat(byeMethod).isNotEmpty();

        // custom assertion
        assertThat(byeMethod.orElseThrow())
                .hasCircuitBreakerAnnotation()
                .doesNotHaveAnyCircuitBreakerAttribute();

        // ...   
}
````

But, I can remove it if you think it's better.

File: pom.xml (hunk 10 above)
Comment (hbelmiro) [reply pom.xml:49]: No. That dependency is used in compile scope only by quarkus-openapi-generator-test-utils, which is then used in test scope by the other modules. The other modules don't have access to assertj in compile scope.

File: deployment/src/test/java/io/quarkiverse/openapi/generator/deployment/assertions/Assertions.java (hunk 8 above)
Comment (fjtirado) [reply deployment/src/test/java/io/quarkiverse/openapi/generator/deployment/assertions/Assertions.java:12]: Thanks for the explanation. Keep it in the way you did.

File: pom.xml (hunk 10 above)
Comment (fjtirado) [reply pom.xml:49]: Ok, thanks for the explanation

**Comment by hbelmiro** [comment]
Thanks for the review @fjtirado.
I committed the requested changes.

**Review by fjtirado — APPROVED** [review]
Thanks @hbelmiro

**Comment by ricardozanini** [comment]
@hbelmiro can you please reabase?

File: deployment/src/test/java/io/quarkiverse/openapi/generator/deployment/circuitbreaker/FileConfigSource.java [hunk 17]
```diff
@@ -0,0 +1,51 @@
...
+import java.io.IOException;
+import java.io.UncheckedIOException;
+import java.util.Map;
+import java.util.Properties;
+import java.util.Set;
+import java.util.stream.Collectors;
+
+import org.eclipse.microprofile.config.spi.ConfigSource;
+
+final class FileConfigSource implements ConfigSource {
```
ricardozanini, line 12: Are you using this `ConfigSource` to parse the application properties file and put the Circuit Breaker properties in the generator context? Smallrye won't have a similar feature?

deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/circuitbreaker/CircuitBreakerConfigurationParser.java
```diff
@@ -0,0 +1,55 @@
...
+
+import java.util.List;
+import java.util.Map;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+import java.util.stream.StreamSupport;
+
+import org.eclipse.microprofile.config.Config;
+
+public final class CircuitBreakerConfigurationParser {
```
ricardozanini, line 11: This is nice! But we need to have a holistic approach to reading these properties (see the authentication feature, we have a very similar way to get the properties). Not sure if we really need this since the `CodegenProvider` already has access to the properties in build time.

ricardozanini (commented):
Excellent work! I'm just worried about the way we consume the properties in general. It should be aligned with what we already have unless we have a strong reason not to.

deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/circuitbreaker/CircuitBreakerConfigurationParser.java
hbelmiro: As we discussed, this is a distinct case. So, we'll keep this way for now.

hbelmiro:
> @hbelmiro can you please reabase?

Done @ricardozanini.

test-utils/pom.xml
```diff
@@ -0,0 +1,33 @@
...
+        <artifactId>quarkus-openapi-generator-parent</artifactId>
+        <groupId>io.quarkiverse.openapi.generator</groupId>
+        <version>1.0.0-SNAPSHOT</version>
+    </parent>
+    <artifactId>quarkus-openapi-generator-test-utils</artifactId>
+    <name>Quarkus - Openapi Generator - Test Utils</name>
+
+    <properties>
+        <version.org.assertj>3.22.0</version.org.assertj>
+        <version.com.github.javaparser>3.24.0</version.com.github.javaparser>
```
ricardozanini, line 16: Do we need these versions here? Also, our bot just updated the parser to 3.24.2.

ricardozanini (approved):

test-utils/pom.xml
hbelmiro: Good catch.
I removed them.

ricardozanini (approved):

ricardozanini:
@all-contributors please add @hbelmiro for code

PR #31, opened 2022-03-28
Fix #28 - Add Enum Outer Class template and values cache

### LINKED ISSUE #28: Wrong class generated for enum
Ginevn the follwogin OpenApi scheme:

```yaml
   ConnectorNamespaceState:
      type: string
      enum:
        - disconnected
        - ready
        - deleting
```

the extension generates an invalid class:

```java
@JsonIgnoreProperties(ignoreUnknown = true)
public enum ConnectorNamespaceState {
    private String value;

    ConnectorNamespaceState(String value){
        this.value = value;
    }

    @Override
    @JsonValue
    public String toString() {
        return String.valueOf(value);
    }

    @JsonCreator
    public static ConnectorNamespaceState fromValue(String text) {
        for (ConnectorNamespaceState b : ConnectorNamespaceState.values()) {
            if (String.valueOf(b.value).equals(text)) {
                return b;
            }
        }
        throw new IllegalArgumentException("Unexpected value '" + text + "'");
    }
}
````

Which is wrong as it does not include the listed enum values.

As an additional note, `ConnectorNamespaceState.values()` should probably be cached as each call to `values()` creates a copy of the enum values for each invocation

### PR DESCRIPTION
**Pull request by ricardozanini**
See #28

### DIFF
--- a/deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/codegen/OpenApiGeneratorCodeGenBase.java
+++ b/deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/codegen/OpenApiGeneratorCodeGenBase.java
@@ -1,7 +1,5 @@
 package io.quarkiverse.openapi.generator.deployment.codegen;
 
-import static io.quarkiverse.openapi.generator.deployment.SpecConfig.API_PKG_SUFFIX;
-import static io.quarkiverse.openapi.generator.deployment.SpecConfig.MODEL_PKG_SUFFIX;
 import static io.quarkiverse.openapi.generator.deployment.SpecConfig.getResolvedBasePackageProperty;
 
 import java.io.IOException;
@@ -26,8 +24,6 @@ public abstract class OpenApiGeneratorCodeGenBase implements CodeGenProvider {
     static final String YML = ".yml";
     static final String JSON = ".json";
 
-    private static final String DEFAULT_PACKAGE = "org.openapi.quarkus";
-
     @Override
     public String inputDirectory() {
         return "openapi";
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

**Comment by ricardozanini** [comment]
Fixes #28

File: deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/codegen/OpenApiGeneratorCodeGenBase.java [hunk 1]
```diff
@@ -47,12 +43,9 @@ public boolean trigger(CodeGenContext context) throws CodeGenException {
...
                             final String basePackage = context.config()
                                     .getOptionalValue(getResolvedBasePackageProperty(openApiFilePath), String.class)
-                                    .orElse(DEFAULT_PACKAGE);
+                                    .orElse("");
                             final OpenApiClientGeneratorWrapper generator = new OpenApiClientGeneratorWrapper(
-                                    openApiFilePath.normalize(), outDir)
-                                            .withApiPackage(basePackage + API_PKG_SUFFIX)
-                                            .withModelPackage(basePackage + MODEL_PKG_SUFFIX)
-                                            .withBasePackage(basePackage);
+                                    openApiFilePath.normalize(), outDir).withBasePackage(basePackage);
```
fjtirado, line 48: How about writing the same behaviour this way
```
 final OpenApiClientGeneratorWrapper generator = new OpenApiClientGeneratorWrapper(openApiFilePath.normalize(), outDir));
context.config().getOptionalValue(getResolvedBasePackageProperty(openApiFilePath), String.class).map(generator::withBasePackage); 

```

fjtirado (approved):

PR #37, opened 2022-03-31
Add support to generate stubs from any InputStream

ricardozanini:
WIP


runtime/src/main/java/io/quarkiverse/openapi/generator/codegen/SpecInputModel.java
```diff
@@ -0,0 +1,51 @@
...
+    private final String filename;
+
+    public SpecInputModel(final String filename, final InputStream inputStream) {
+        requireNonNull(inputStream, "InputStream can't be null");
+        requireNonNull(filename, "File name can't be null");
+        this.inputStream = inputStream;
+        this.filename = filename;
+    }
+
+    public String getFileName() {
```
fjtirado, line 20: why not uri rather than filename?

deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/codegen/OpenApiGeneratorStreamCodeGen.java
```diff
@@ -0,0 +1,104 @@
...
+    @Override
+    public boolean trigger(CodeGenContext context) throws CodeGenException {
+        final Path outDir = context.outDir();
+        boolean generated = false;
+
+        for (final OpenApiSpecInputProvider provider : this.providers) {
+            for (SpecInputModel inputModel : provider.read()) {
+                if (inputModel == null) {
+                    throw new CodeGenException("SpecInputModel from provider " + provider + " is null");
+                }
```
fjtirado, line 67: Im not sure if this is the proper flow
I mean, you are reading the input stream and copying it to a local file, is that really needed? 
Does not the generator directly accepts the input stream to generate the content?

runtime/src/main/java/io/quarkiverse/openapi/generator/codegen/SpecInputModel.java
ricardozanini: We just need the file name from the origin, which will be used to save it locally. So if we have:\

- http://example.com/myspecA.json -> myspecA.json
- /home/luke/files/myspecB.yaml -> myspecB.yaml
- classpath://files/myspecC.yml -> myspecC.yaml

URI won't be used in this extension since our input model is a stream.

fjtirado: How are we going to map a function which operation is  /home/luke/files/myspecB.yaml#doSomething to the classes generated from that URI is the information that we pass to the generation is just myspecA.json?
The answer is by defining a known mapping between URI and file in workflow. That mapping might be tricky 
Imagine we have these two URIS: http://example1.com/myspecA.json and  http://example2.com/myspecA.json, we need to have the whole URI into account when generating the file name. Anyway, this is an issue for the implementor, not this PR

PR #45, opened 2022-04-05
Fix #38 - Add option to skip generation of deprecated attributes

### LINKED ISSUE #38: Handling of deprectated fields
If a schema contains a deprecated field, the generated class may be invalid, as example given the schema below: 

```yaml
    MetaV1Condition:
      type: object
      properties:
        type:
          type: string
        reason:
          type: string
        message:
          type: string
        status:
          type: string
        lastTransitionTime:
          type: string
          deprecated: true
        last_transition_time:
          type: string
          #format: date-time
```

The generated class is:

```java
@JsonIgnoreProperties(ignoreUnknown = true)
public class MetaV1Condition  {

    private String type;
    private String reason;
    private String message;
    private String status;
    private String lastTransitionTime;
    private String lastTransitionTime;
}
 ```
 
As you can notice, `lastTransitionTime` get duplicated which result in a complication error. 
 
It would be nice to have an option to control the behavior of the generator i.e. to skip any deprecated fields.

### PR DESCRIPTION
**Pull request by ricardozanini**
Fixes #38 

- [x] Add more test cases
- [x] Write Docs
- [x] Clean up

### CODE REVIEW

File: deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/template/OpenApiNamespaceResolver.java [hunk 1]
```diff
@@ -0,0 +1,59 @@
+package io.quarkiverse.openapi.generator.deployment.template;
```
fjtirado, line 1: as you know, This file is not longer needed, you can add the getDeprecatedModelAttr to  https://github.com/quarkiverse/quarkus-openapi-generator/blob/main/deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/template/QuteTemplatingExtension.java

deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/template/QuteTemplatingEngineAdapter.java
```diff
@@ -36,6 +36,7 @@ public QuteTemplatingEngineAdapter() {
         this.engine = Engine.builder()
                 .addDefaults()
                 .addValueResolver(new ReflectionValueResolver())
+                .addNamespaceResolver(OpenApiNamespaceResolver.INSTANCE)
```
fjtirado, line 39: Same, already present in latest main branch

deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/wrapper/OpenApiClientGeneratorWrapper.java
```diff
@@ -76,7 +76,16 @@ public OpenApiClientGeneratorWrapper withBasePackage(final String pkg) {
      * @return this wrapper
      */
     public OpenApiClientGeneratorWrapper withCircuitBreakerConfiguration(final Map<String, List<String>> config) {
-        configurator.addAdditionalProperty("circuit-breaker", config);
+        if (config != null) {
```
fjtirado, line 79: when will config be null?

deployment/src/main/resources/templates/pojo.qute
```diff
@@ -32,6 +32,7 @@ import com.fasterxml.jackson.annotation.JsonProperty;
 public class {m.classname} {#if m.parent}extends {m.parent}{/if}{#if m.serializableModel} implements Serializable{/if} {
 
     {#for v in m.vars}
+    {#if !v.deprecated || openapi:genDeprecatedModelAttr(package, m.classname, codegen)}
```
fjtirado, line 35: Im wondering it it wont be better to have 
v.isDeprecated( ) method, so not using namespace method but template extension https://quarkus.io/guides/qute-reference#template_extension_methods
Yes, it will take a little bit more of investigation (to see how to configure that in the template engine without annotations) but the template will be far simple.

fjtirado (commented):
Some comments

deployment/src/main/resources/templates/pojo.qute
ricardozanini: Actually, I think the former approach was better when I implemented the `ValueResolver`. This way we won't need to send parameters. I'd rather have a namespaced extension available in the `openapi` namespace. This way there's a clear separation of concerns of the functions we are using there.

deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/wrapper/OpenApiClientGeneratorWrapper.java
ricardozanini: It's just a guard check.

deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/template/OpenApiNamespaceResolver.java
ricardozanini: Actually, I'll keep the naming  `OpenApiNamespaceResolver` since it has a clear definition of what it is.

fjtirado (approved):

PR #54, opened 2022-04-28
Adding custom properties to OpenApiSpecInputProvider

ricardozanini:
In this PR we add the possibility to configure the generated code via the `OpenApiSpecInputProvider` interface.

### DIFF
--- a/deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/SpecConfig.java
+++ b/deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/SpecConfig.java
@@ -1,23 +1,21 @@
 package io.quarkiverse.openapi.generator.deployment;
 
 import java.nio.file.Path;
-import java.util.Collections;
 import java.util.Map;
 
-import io.quarkus.runtime.annotations.ConfigGroup;
 import io.quarkus.runtime.annotations.ConfigItem;
+import io.quarkus.runtime.annotations.ConfigPhase;
 import io.quarkus.runtime.annotations.ConfigRoot;
 
 // This configuration is read in codegen phase (before build time), the annotation is for document purposes and avoiding quarkus warns
-@ConfigGroup
-@ConfigRoot(prefix = SpecConfig.BUILD_TIME_CONFIG_PREFIX)
+@ConfigRoot(name = SpecConfig.BUILD_TIME_CONFIG_PREFIX, phase = ConfigPhase.BUILD_TIME)
 public class SpecConfig {
 
-    public static final String BUILD_TIME_CONFIG_PREFIX = "quarkus.openapi-generator.codegen";
+    static final String BUILD_TIME_CONFIG_PREFIX = "openapi-generator.codegen";
     public static final String API_PKG_SUFFIX = ".api";
     public static final String MODEL_PKG_SUFFIX = ".model";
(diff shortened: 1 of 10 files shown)


deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/SpecConfig.java
```diff
@@ -1,35 +1,29 @@
...
     /**
      * OpenAPI Spec details for codegen configuration.
      */
     @ConfigItem(name = "spec")
-    Map<String, SpecItemConfig> specItem;
-
-    public Map<String, SpecItemConfig> getSpecItem() {
-        return Collections.unmodifiableMap(specItem);
-    }
+    public Map<String, SpecItemConfig> specItem;
```
fjtirado, line 26: If this public field ever used?, I cannot find any code using it

deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/codegen/SpecInputModel.java
```diff
@@ -3,12 +3,21 @@
...
+import org.eclipse.microprofile.config.spi.ConfigSource;
+
+import io.quarkiverse.openapi.generator.deployment.SpecConfig;
+import io.smallrye.config.PropertiesConfigSource;
+
 public class SpecInputModel {
 
     private final InputStream inputStream;
     private final String filename;
+    private final Map<String, String> codegenProperties = new HashMap<>();
```
fjtirado, line 20: Should not this constructor call the new one using this reserved word?, passing null as the default package name

deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/SpecItemConfig.java
```diff
@@ -1,14 +1,24 @@
 package io.quarkiverse.openapi.generator.deployment;
```
fjtirado, line 1: I do not think this class is needed any longer. 
I just try deleting it and the mvn clean install still works

deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/SpecConfig.java
ricardozanini: It's for document purposes only and Quarkus bind this property, so it must be public.

deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/SpecItemConfig.java
ricardozanini: It's there for doc config / quarkus bind purposes only. In your use case you won't see the warning because you don't have a property in your env prefixing with "quarkus.openapi-generator.codegen.spec", you won't see it, but end users will.

deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/codegen/SpecInputModel.java
ricardozanini: Why? What's the difference? Every time this class is instantiated, this field will be instantiated. If you mean the second constructor added to this class, yes. I copied and pasted, I will remove and call this(param, param). Forgot to change.

PR #68, opened 2022-05-13
Add token propagation support

wmedvede:


wmedvede:
Hello @ricardozanini would you mind review this PR while I continue working on the tests and documentation updates?

p.s. tests are in progress, so don't spend time on them.

Thanks!

**Description:**

This PR introduces the ability to enable the security token propagation for particular securityScheme present on the OpenApi specification files, and additional related configurations.

The intention is to facilitate these configurations by using Quarkus styled properties naming, at the same time the configurations can be done per specification file and per securityScheme on that file.

Given an OpenApi specification file **petstore.json** the prefix **quarkus.openapi-generator.petstore_json** must be used to configure particular settings for that file.

Below we can see some example configurations assuming that we have the following securitySchemes defined on the file. (the types where included on the names for simplicity)

  "**example-http-bearer**"
  "**example-http-basic**"
  "**example-oauth2**"
  "**example-api-ke**y"


```
# Bearer token configuration example for the "example-http-bearer" security scheme.
quarkus.openapi-generator.petstore_json.auth.example_http_bearer.bearer-token=ABCDEF

# Bearer token propagation example for the "example-http-bearer" security scheme.
# Configures the token propagation and the header attribute from where to read the token to be propagated.
quarkus.openapi-generator.petstore_json.auth.example_http_bearer.token-propagation=true
quarkus.openapi-generator.petstore_json.auth.example_http_bearer.header-name=User-Custom-Header

# Basic authentication example configuration for the "example-http-basic" securityScheme
quarkus.openapi-generator.petstore_json.auth.example_http_basic.username=USER1
quarkus.openapi-generator.petstore_json.auth.example_http_basic.password=USER1PASSWORD

# Oauth2 token propagation configuration example for the "example-oauth2" security scheme.
quarkus.openapi-generator.petstore_json.auth.example_oauth2.token-propagation=true

# ApiKey configuration example for the "example-api-key" security scheme.
quarkus.openapi-generator.petstore_json.auth.example_api_key.api-key=VWXYZ
```

deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/CodegenBuildTimeConfig.java
```diff
@@ -9,8 +9,8 @@
 import io.smallrye.config.common.utils.StringUtil;
 
 // This configuration is read in codegen phase (before build time), the annotation is for document purposes and avoiding quarkus warns
-@ConfigRoot(name = CodegenConfig.CODEGEN_TIME_CONFIG_PREFIX, phase = ConfigPhase.BUILD_TIME)
-public class CodegenConfig {
+@ConfigRoot(name = CodegenBuildTimeConfig.CODEGEN_TIME_CONFIG_PREFIX, phase = ConfigPhase.BUILD_TIME)
```
ricardozanini, line 12: Walter, this config is related to Codegen only. Build time happens after codegen. Maybe we need to review the prefixes:

1. `quarkus.openapi-generator.codegen`
2. `quarkus.openapi-generator.buildtime`
3. `quarkus.openapi-generator.runtime`

wdyt?

ricardozanini (commented):
I'll finish the review later.

deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/CodegenBuildTimeConfig.java
wmedvede: Hi @ricardozanini , I see what you mean but this was my reasoning:

AFAIK in Quarkus, we have: three config phases that we can use for our needs, BUILD_TIME, BUILD_AND_RUN_TIME_FIXED and RUN_TIME. All right, so with this in mind, independently on the class name, I don't believe we have the ability to configure something in a kind "codegen" phase. 
If I'm not wrong, ok, we managed to generated the code before some other tasks, but we are still in the Quarkus BUILD_TIME phase. I am right?

The use of the  "CodegenBuildTimeConfig" comes from a kind of Quarkus naming style (but not mandatory) so at the class name level we indicate that this class corresponds to BUILD_TIME configurations, that in our case, this somehow means "codegen" that in the end is what we do.
So this class preserves let's say the config prefix we had: "openapi-generator.codegen", and I believe it's fine since users will understand that this is related with the code generation we are producing.

Then we have configurations that apply for the runtime, basically the new part I've added plus some small refactoring of some already existing runtime related configs.
For these part,  we have directly the prefix "openapi-generator", It's shorter, and I think it makes sense, not a big deal to change, really a mater of definition.

And that was my reasoning.

Regarding your suggestion, I think that 

1) quarkus.openapi-generator.codegen is good!, in the end it's what we do.

2) quarkus.openapi-generator.buildtime not sure, we don't need by now. I'd believe we can introduce it if we need at some point.

3) quarkus.openapi-generator.runtime and here, I don't know, in general around Quarkus I believe that property names don't indicate "runtime" or "buildtime" instead you discover the nature of the property by reading the documentation. 

wdyt?

fjtirado (approved):

wmedvede:
> I'll finish the review later.

no worries, btw, the tests related to the providers are in. Let's finish the discussion see how to continue.

deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/CodegenBuildTimeConfig.java
ricardozanini: Hi @wmedvede!

> If I'm not wrong, ok, we managed to generated the code before some other tasks, but we are still in the Quarkus BUILD_TIME phase. I am right?

No. The "codegen" phase is not a build phase, it's tied to the Quarkus Maven Plugin "generate-code" execution phase. So it's not part of the Quarkus buildtime. That's why I'd rather keep a separated config class just for documentation/utils purpose. 

This class is there to not make Quarkus WARN the console log all the time. It's not read by the build nor runtime phases. Our `CodeGen` interface implementation read the properties via the `config` parameter, which is a generic accessor. 

> So this class preserves let's say the config prefix we had: "openapi-generator.codegen", and I believe it's fine since users will understand that this is related with the code generation we are producing.

It's better to have another class that holds the build time configuration if that makes sense. But I believe we only have "codegen" and "runtime" configuration. So to avoid misinterpretation, please remove the "BuildTime" prefix from this classes.

> Then we have configurations that apply for the runtime, basically the new part I've added plus some small refactoring of some already existing runtime related configs.

That's perfect! Runtime configuration should be handled the way you suggested here.

> quarkus.openapi-generator.buildtime not sure, we don't need by now. I'd believe we can introduce it if we need at some point.

We don't need this now.

> quarkus.openapi-generator.runtime and here, I don't know, in general around Quarkus I believe that property names don't indicate "runtime" or "buildtime" instead you discover the nature of the property by reading the documentation.

Well, we can think on another prefix if that makes sense. But can you test if `quarkus.openapi-generator` (no prefix, meaning for runtime) and `quarkus.openapi-generator.codegen` won't clash? If not, we can go this way!

runtime/src/main/java/io/quarkiverse/openapi/generator/CodegenConfig.java
```diff
@@ -0,0 +1,43 @@
...
+import io.quarkus.runtime.annotations.ConfigItem;
+import io.quarkus.runtime.annotations.ConfigPhase;
+import io.quarkus.runtime.annotations.ConfigRoot;
+import io.smallrye.config.common.utils.StringUtil;
+
+/**
+ * This class represents the runtime configurations for the openapi-generator extension.
+ */
+@ConfigRoot(name = CodegenConfig.RUNTIME_TIME_CONFIG_PREFIX, phase = ConfigPhase.RUN_TIME)
+public class CodegenConfig {
```
ricardozanini, line 15: This class must be renamed (maybe `OpenApiGeneratorConfig`). There's no codegen config in runtime.

runtime/src/main/java/io/quarkiverse/openapi/generator/SpecItemAuthConfig.java
```diff
@@ -0,0 +1,94 @@
...
+import io.quarkiverse.openapi.generator.providers.BearerAuthenticationProvider;
+import io.quarkus.runtime.annotations.ConfigGroup;
+import io.quarkus.runtime.annotations.ConfigItem;
+
+/**
+ * This class represents the runtime authentication related configuration for an individual securityScheme present
+ * on an OpenApi spec definition, i.e. the provided files.
+ */
+@ConfigGroup
+public class SpecItemAuthConfig {
```
ricardozanini, line 17: Can we keep `AuthConfig`?

ricardozanini (commented):
Let's settle the naming conflict of the configuration classes. Many thanks, Walter!

wmedvede:
Hi  @ricardozanini , all the suggestions were incorporated, take a look please and let me know if we have something else to improve/modify. 
Thanks!

wmedvede:
HI @ricardozanini the suggested class renaming to AbstractAuthenticationPropagationHeadersFactory and AuthenticationPropagationHeadersFactory was added.

If there's any other remaining suggestion let me know,
Thanks

PR #78, opened 2022-05-30
Add token propagation support

wmedvede:
- documentation updates


deployment/pom.xml
```diff
@@ -25,6 +25,10 @@
       <groupId>io.quarkus</groupId>
       <artifactId>quarkus-rest-client-jackson-deployment</artifactId>
     </dependency>
+    <dependency>
+      <groupId>io.quarkus</groupId>
+      <artifactId>quarkus-oidc-client-filter-deployment</artifactId>
```
wmedvede, line 30: @ricardozanini  for the record, this leaves the extension prepared to work
1) with the client filter by not having to add any additional dependencies on the users
2) work well in native mode

runtime/pom.xml
```diff
@@ -26,7 +26,6 @@
     <dependency>
       <groupId>io.quarkus</groupId>
       <artifactId>quarkus-oidc-client-filter</artifactId>
-      <scope>provided</scope>
```
wmedvede, line 29: @ricardozanini  for the record, this leaves the extension prepared to work
1) with the client filter by not having to add any additional dependencies on the users
2) work well in native mode

wmedvede:
Hi  @ricardozanini , this PR includes the documentation updates for the token propagation plus a couple of fixes for the native mode. Would you mind review?

Thanks!

deployment/pom.xml
ricardozanini: But it does bring the dependency if not needed in scenarios where users won't use OAuth2?

runtime/pom.xml
ricardozanini: I'd rather not bring an optional dependency if users won't use OAuth2. :(

ricardozanini:
The OIDC dependency was added because of native?

README.md
```diff
@@ -213,17 +214,92 @@ quarkus.oidc-client.petstore_auth.grant-options.password.password=alice
 quarkus.oidc-client.petstore_auth.client-id=petstore-app
 ```
 
-The configuration suffix `quarkus.oidc-client.petstore_auth` is exclusive for the schema defined in the specification file.
+The configuration suffix `quarkus.oidc-client.petstore_auth` is exclusive for the schema defined in the specification file and the schemaName is sanitized by aplying the rules described above.
```
ricardozanini, line 217:
```suggestion
The configuration suffix `quarkus.oidc-client.petstore_auth` is exclusive for the schema defined in the specification file and the `schemaName` is sanitized by applying the rules described above.
```

README.md
```diff
@@ -213,17 +214,92 @@ quarkus.oidc-client.petstore_auth.grant-options.password.password=alice
...
 
-For this to work you **must** add [Quarkus OIDC Client Filter Extension](https://quarkus.io/guides/security-openid-connect-client#oidc-client-filter) to your project:
```
ricardozanini, line 218: @wmedvede if it's not mandatory to native compilation, I'd rather keep this optional, so we won't bring unnecessary dependencies to users' projects.

README.md
```diff
@@ -213,17 +214,92 @@ quarkus.oidc-client.petstore_auth.grant-options.password.password=alice
...
+When configured, you can propagate the authorization tokens passed to your own service along the invocations to the rest clients generated by the quarkus-openapi-generator.
 
-<dependency>
-  <groupId>io.quarkus</groupId>
-  <artifactId>quarkus-oidc-client-filter</artifactId>
-</dependency>
-````
+Let's see how it works by following a simple example:
+
+Imagine that we have an `"updatePet"` operation, that is  defined in the `petstore.json` specification file, and that is secured with the `"petstore_auth"` security scheme.
```
ricardozanini, line 226:
```suggestion
Imagine that we have a `updatePet` operation defined in the `petstore.json` specification file and secured with the `petstore_auth` security scheme.
```

README.md
```diff
@@ -213,17 +214,92 @@ quarkus.oidc-client.petstore_auth.grant-options.password.password=alice
...
 
-The configuration suffix `quarkus.oidc-client.petstore_auth` is exclusive for the schema defined in the specification file.
+The configuration suffix `quarkus.oidc-client.petstore_auth` is exclusive for the schema defined in the specification file and the schemaName is sanitized by aplying the rules described above.
 
-For this to work you **must** add [Quarkus OIDC Client Filter Extension](https://quarkus.io/guides/security-openid-connect-client#oidc-client-filter) to your project:
+## Authorization Token Propagation
 
-````xml
+The authorization token propagation can be used with OpenApi operations that are secured with a security scheme of type "oauth2" or "bearer".
+When configured, you can propagate the authorization tokens passed to your own service along the invocations to the rest clients generated by the quarkus-openapi-generator.
```
ricardozanini, line 222:
```suggestion
The authorization token propagation can be used with OpenApi operations secured with a security scheme of type "oauth2" or "bearer".
When configured, you can propagate the authorization tokens passed to your service and the invocations to the REST clients generated by the quarkus-openapi-generator.
```

README.md
```diff
@@ -213,17 +214,92 @@ quarkus.oidc-client.petstore_auth.grant-options.password.password=alice
...
+    // Set the corresponding values to the Pet instance.
+  }
+}
+```
+
+Let's see what happens when the PetResource service `customUpdatePet` operation is invoked by a third party.
+
+### Default flow
+1) The `customUpdatePet` operation is invoked.
+2) An authorization token is obtained by using the corresponding `"petstore_auth"` OidcClient configuration. (for more information see [OAuth2 Authentication](#oauth2-authentication))
```
ricardozanini, line 285:
```suggestion
2) An authorization token is obtained using the corresponding `petstore_auth` OidcClient configuration. (for more information see [OAuth2 Authentication](#oauth2-authentication))
```

README.md
```diff
@@ -213,17 +214,92 @@ quarkus.oidc-client.petstore_auth.grant-options.password.password=alice
...
+3) The authorization token is automatically passed along the PetApi `updatePet` operation execution by using an automatically generated request filter, etc.
+
+### Propagation flow
+However, there are scenarios where we simply want to propagate the authorization token that was originally passed to the PetResource service when the `customUpdatePet` operation was invoked, instead of having to obtain it by using the OidcClient.
+
+1) The user service `customUpdatePet` operation is invoked, and an authorization token is passed by the third party typically by using the http `"Authorization"` header.
```
ricardozanini, line 291:
```suggestion
1) The user service `customUpdatePet` operation is invoked, and an authorization token is passed by the third party typically by using the HTTP `Authorization` header.
```

README.md
```diff
@@ -213,17 +214,92 @@ quarkus.oidc-client.petstore_auth.grant-options.password.password=alice
...
+2) The incoming authorization token is automatically passed along the PetApi `updatePet` operation execution according to the user provided configuration.
+
+> **⚠️** When configured, the token propagation apply to all the operations secured with the same securityScheme in the same specification file.
```
ricardozanini, line 294:
```suggestion
> **⚠️** When configured, the token propagation applies to all the operations secured with the same `securityScheme` in the same specification file. 
```

wmedvede:
> The OIDC dependency was added because of native?

yes, by adding it we have native mode working, but also users doesn't have to add the
```
     <dependency>
      <groupId>io.quarkus</groupId>
      <artifactId>quarkus-oidc-client-filter</artifactId>
    </dependency>
```
when they are working with the extension.
Otherwise they must be conscious of the fact that a particular security scheme has "oauth2" type and add it manually.

deployment/pom.xml
wmedvede: somehow yes, but I believe it's a fair trade-off between letting the extension be more "self-contained" or shift some configuration aspects to the user. Also, when we go to native I'm not sure if we could introduce other issues.

runtime/pom.xml
wmedvede: Honestly I don't see a huge problem with that, I fact, this is something happens many times with other let's say more high-level extensions that behind the scene they use other extensions.

wmedvede:
> Looks great! Thank you for writing all these docs. The only thing I'm concerned about is that we are now enforcing the whole OIDC dependencies.

I'll do a quick check to see how this might affect end users + native mode...

ricardozanini:
> Otherwise they must be conscious of the fact that a particular security scheme has "oauth2" type and add it manually

That's what we state right now. If their spec requires oauth2, they add the dependency. Since this is meant for cloud applications, not sure if bringing more dependencies will favor the experience.

README.md
wmedvede: this was restored

ricardozanini (approved):

Issue #86, opened 2022-06-22
@ApplicationScoped required for mocking api and missing 


miguelchico:
Hello

Following the official Quarkus guide (https://quarkus.io/blog/mocking/#using-injectmock) I found out I need to use the @ApplicationScoped annotation in the interface defining the api and that annotation is missing in the generated code here.

Is there any way of adding it? Is it possible mocking in any other way?

Thanks

ricardozanini:
Hi @miguelchico! Currently, you can't. We can add it in here, though: https://github.com/quarkiverse/quarkus-openapi-generator/blob/main/deployment/src/main/resources/templates/api.qute#L40.

Can you send a PR and run some tests to evaluate it? 

As a workaround, I think you can try overwriting the interface and adding your annotation there.

miguelchico:
@ricardozanini I sent a super small PR, please check it when you have the time.

Thank you

Shohou:
@ricardozanini, @miguelchico is this annotation still needed with current quarkus?
https://quarkus.io/guides/rest-client-reactive#mocking-with-injectmock - doesn't require it...
I have opposite situation. I implement generated Api interface with my mock class and because of `ApplicationScoped` annotation quarkus thinks it's server-side rest methods and application exposes them for calling from outside.

Issue #85, opened 2022-06-18
How to use additionalModelTypeAnnotations


miguelchico:
I would like to include some extra additional annotations to the models generated here.
I come from openapi-generator project and there is a way of doing it using a property called additionalModelTypeAnnotations.

I’ve done a bit of digging and I found same property mentioned on templates here as well but I wasn’t able to find the way of adding configuration for it.

Is it possible? If it is not I think that would be a good addition to the project.

Thank you

PR #87, opened 2022-06-23
Add @ApplicationScoped to generated api to allow mocking api for testing

miguelchico:
PR to fix issue #86.
Adding @ApplicationScoped annotation to generated API will allow us to mock it in unit test according to official Quarkus guidelines https://quarkus.io/blog/mocking/#using-injectmock

### DIFF
--- a/deployment/src/main/resources/templates/api.qute
+++ b/deployment/src/main/resources/templates/api.qute
@@ -20,6 +20,8 @@ import javax.ws.rs.*;
 import javax.ws.rs.core.Response;
 import javax.ws.rs.core.MediaType;
 
+import javax.enterprise.context.ApplicationScoped;
+
 import io.quarkiverse.openapi.generator.annotations.GeneratedClass;
 import io.quarkiverse.openapi.generator.annotations.GeneratedMethod;
 import io.quarkiverse.openapi.generator.annotations.GeneratedParam;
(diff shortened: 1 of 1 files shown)


ricardozanini:
@miguelchico were you able to test this approach in your application? I mean, with this PR you're able to mock the generated stub? 

Another way to achieve mocking is to use Wiremock and change the endpoint to the mocked server.

I'm just wondering if this annotation won't bring any side effects to the user's project.

miguelchico:
Hi @ricardozanini 
I tested my application with the extra annotation and I couldn't find any side effects.
The idea here is to be able to mock that client in a more pure unit test rather than an integration test with a real webserver (mocked with wiremock)

ricardozanini:
@miguelchico I'll prepare a new release since I understand this is quite important for you?

ricardozanini:
@all-contributors please add @miguelchico for code

ricardozanini:
@miguelchico do you think we can wait for #85 then?

miguelchico:
@ricardozanini give me some time in order to do just one new release with both if you are ok, I'll need both anyway so it might be easier for both

ricardozanini:
Yes, that's what I was thinking. Let's include both in the new release. Take your time, please. Thanks for the contribution!

PR #89, opened 2022-06-23
Add capability to use additionalModelTypeAnnotations

### LINKED ISSUE #85: How to use additionalModelTypeAnnotations

miguelchico:
Initial work to fix #85


deployment/src/main/resources/templates/additionalModelTypeAnnotations.qute
```diff
@@ -1 +1,3 @@
-{#for annon in m.additionalModelTypeAnnotations.orEmpty}{annon}{/for}
```
miguelchico, line 1: I'm not an expert in using Qute so I'm not sure if this change can be done in a different way, I'm open to suggestions

ricardozanini (approved):

ricardozanini:
@fjtirado can you please take a look? I'm planning to merge this on Monday and release a new version for @miguelchico 

@miguelchico I'm assuming this PR is ready? Looks ready to me. :)

miguelchico:
@ricardozanini it is ready in my opinion. It would be great to get the new version next week. Thank you!

deployment/src/main/resources/templates/additionalModelTypeAnnotations.qute
fjtirado: Me enither, but Im guessing what happens if you try with 
{additionalModelTypeAnnotations.orEmpty}

ricardozanini:
@miguelchico congrats! https://github.com/quarkiverse/quarkus-openapi-generator/releases/tag/0.8.0

Please take a few hours for Maven Central to update the artifacts.

PR #90, opened 2022-06-23
KOGITO-6458: Define default authentication method to be used by OpenAPI

martinweiler:
The idea of this enhancement is to provide a configuration option that allows users to enforce security for all endpoints, even if the specification file is not explicitly defining these via the `security` element. Example: Assuming we have a spec file that contains the `securitySchemes` definition, but no `security` reference:

```
  securitySchemes:
    basic_auth:
      type: http
      scheme: basic
```

With this enhancement, the generator will create authentication bindings if the following property is defined:

```
quarkus.openapi-generator.codegen.default.security.scheme=basic_auth
```

**Note:** The property value needs to match the name of the `securityScheme` in the spec file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,6 +147,12 @@ pattern: `quarkus.openapi-generator.[filename].[security_scheme_name].auth.[auth
 
 > Tip: on production environments you will likely to use [HashiCorp Vault](https://quarkiverse.github.io/quarkiverse-docs/quarkus-vault/dev/index.html) or [Kubernetes Secrets](https://kubernetes.io/docs/concepts/configuration/secret/) to provide this information for your application.
 
+If your OpenAPI specification file has `securitySchemes` definitions, but no [Security Requirement Object](https://spec.openapis.org/oas/v3.1.0#security-requirement-object) definitions, the generator can create these by default if the following property is defined:
+
+| Description          | Property Key                                                   | Value                                          |
+| -------------------- | -------------------------------------------------------------- | ---------------------------------------------------- |
+| Create security for the referenced security scheme | `quarkus.openapi-generator.codegen.default.security.scheme` | Name of the security scheme object definition, eg. `api_key` or `basic_auth` in the security scheme definitions above |
+
 ### Basic HTTP Authentication
 
 For Basic HTTP Authentication, these are the supported configurations:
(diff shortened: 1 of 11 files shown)


README.md
```diff
@@ -147,6 +147,12 @@ pattern: `quarkus.openapi-generator.[filename].[security_scheme_name].auth.[auth
 
 > Tip: on production environments you will likely to use [HashiCorp Vault](https://quarkiverse.github.io/quarkiverse-docs/quarkus-vault/dev/index.html) or [Kubernetes Secrets](https://kubernetes.io/docs/concepts/configuration/secret/) to provide this information for your application.
 
+If your OpenAPI specification file has `securitySchemes` definitions, but no [Security Requirement Object](https://spec.openapis.org/oas/v3.1.0#security-requirement-object) definitions, the generator can create these by default if the following property is defined:
```
ricardozanini, line 150: I think we can expand a little bit more here and add that all the operations without a security requirement, will now have the default one.
Also, the default one must be defined in the `securitySchemes` list.

ricardozanini (commented):
LGTM! Many thanks for this work! Can I ask to add an integration test based on: https://github.com/quarkiverse/quarkus-openapi-generator/blob/main/integration-tests/example-project/src/test/java/io/quarkiverse/openapi/generator/it/OpenWeatherTest.java

You can copy the spec, remove the sec from the operation, set this config, and do the same verification. This way we guarantee that the apikey is being added to the request. You can even use the same mock.

martinweiler:
@ricardozanini I have added an integration test and updated the README file to expand more on this new functionality.

ricardozanini (approved):
LGTM! Many thanks for this contribution!

deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/wrapper/QuarkusJavaClientCodegen.java
```diff
@@ -78,6 +78,8 @@ public void preprocessOpenAPI(OpenAPI openAPI) {
         super.preprocessOpenAPI(openAPI);
         // add the default server url to the context
         additionalProperties.put(DEFAULT_SERVER_URL, URLPathUtils.getServerURL(this.openAPI, serverVariableOverrides()));
+        additionalProperties.put(OpenApiClientGeneratorWrapper.DEFAULT_SECURITY_SCHEME,
+                GlobalSettings.getProperty(OpenApiClientGeneratorWrapper.DEFAULT_SECURITY_SCHEME, null));
```
fjtirado, line 82: if getProperty returns null, sohuld we add null to the additionalProperties map?

fjtirado (approved):
Just a picky comment about inserting null value into the properties map

deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/wrapper/QuarkusJavaClientCodegen.java
martinweiler: as discussed, changed the getProperty call to not define a default value

PR #94, opened 2022-06-28
Using full path as config key

fjtirado:
For files generated from streams, relative spec file path rather than file name should be used for generating config key

### DIFF
--- a/deployment/src/test/java/io/quarkiverse/openapi/generator/deployment/CodegenConfigTest.java
+++ b/deployment/src/test/java/io/quarkiverse/openapi/generator/deployment/CodegenConfigTest.java
@@ -10,10 +10,17 @@ import org.junit.jupiter.api.Test;
 class CodegenConfigTest {
 
     @Test
-    void verifySpaceEncoding() {
+    void verifyStreamUri() {
         final String resolvedPrefix = CodegenConfig
                 .getBuildTimeSpecPropertyPrefix(Path.of("/home/myuser/open-api-stream/luke/my test openapi.json"));
         assertEquals(String.format(BUILD_TIME_SPEC_PREFIX_FORMAT, "luke_my_test_openapi_json"), resolvedPrefix);
+    }
+
+    @Test
+    void verifySpaceEncoding() {
+        final String resolvedPrefix = CodegenConfig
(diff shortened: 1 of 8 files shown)


deployment/src/test/java/io/quarkiverse/openapi/generator/deployment/CodegenConfigTest.java
```diff
@@ -12,14 +12,14 @@ class CodegenConfigTest {
     @Test
     void verifySpaceEncoding() {
         final String resolvedPrefix = CodegenConfig
-                .getBuildTimeSpecPropertyPrefix(Path.of("/home/myuser/luke/my test openapi.json"));
-        assertEquals(resolvedPrefix, String.format(BUILD_TIME_SPEC_PREFIX_FORMAT, "my_test_openapi_json"));
+                .getBuildTimeSpecPropertyPrefix(Path.of("/home/myuser/open-api-stream/luke/my test openapi.json"));
+        assertEquals(String.format(BUILD_TIME_SPEC_PREFIX_FORMAT, "luke_my_test_openapi_json"), resolvedPrefix);
```
ricardozanini, line 16: If you kept the `/home/myuser/luke/my test openapi.json` so the result would be `home_myuser_luke_my_test_openapi_json`? Please keep both tests. You can add a new one verifying the use case where the root path is one of the expected.

ricardozanini (commented):
Can you please also update the README? Specifically here: 

> "Now, create the directory openapi under your src/main/ path and add the OpenAPI spec files there. We support JSON, YAML and YML extensions."

One more thing, please update one of the integration tests to have the spec under `/src/main/resources/openapi/specs/myspec.yaml` and check if the property is correctly defined as `specs_myspec_yaml`.

Thanks!

deployment/src/test/java/io/quarkiverse/openapi/generator/deployment/CodegenConfigTest.java
fjtirado: I restore the previous test and added a new one. 
The idea is that for files generated under $OUTPUTPATH/open-api-stream, the config key will be the relative path (previous it was the file name)
I do not understand how is this related with the readme file.

fjtirado:
> Can you please also update the README? Specifically here:

This just fixes a bug, does not affect the readme

ricardozanini (approved):

PR #95, opened 2022-06-28
Do not put a default url if there is not server url

fjtirado:
The goal of this PR is that if configuration is missing, the following message is printed

`Caused by: java.lang.IllegalArgumentException: Unable to determine the proper baseUrl/baseUri. Consider registering using @RegisterRestClient(baseUri="someuri"), @RegisterRestClient(configKey="orkey"), or by adding 'quarkus.rest-client."subtration_spec_yaml".url' or 'quarkus.rest-client."subtration_spec_yaml".uri' to your Quarkus configuration`


ricardozanini (approved):

ricardozanini:
Can you please add a test to verify if the annotation in the generated stub doesn't have a `baseUrl` defined?

ricardozanini (approved):

PR #102, opened 2022-07-07
Removed import that might cause compilation error due to ambiguity

hbelmiro:
This PR removes `import javax.ws.rs.core.Response;` from the api.qute template.

Maybe, as a follow up improvement, we should also [remove all imports from all the templates and qualify the classes on their usages](https://github.com/quarkiverse/quarkus-openapi-generator/issues/103).
WDYT @ricardozanini?

### DIFF
--- a/deployment/src/main/resources/templates/api.qute
+++ b/deployment/src/main/resources/templates/api.qute
@@ -17,7 +17,6 @@ import java.io.OutputStream;
 import java.util.List;
 import java.util.Map;
 import javax.ws.rs.*;
-import javax.ws.rs.core.Response;
 import javax.ws.rs.core.MediaType;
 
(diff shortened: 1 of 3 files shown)


ricardozanini:
> 

I think it's a good idea. Can you elaborate a little bit more in an issue, so we can have it documented?

hbelmiro:
> I think it's a good idea. Can you elaborate a little bit more in an issue, so we can have it documented?

Issue created: https://github.com/quarkiverse/quarkus-openapi-generator/issues/103

hbelmiro:
@ricardozanini the PR description was not clear and might led you to misunderstand it.
Sorry for that.
I updated the PR description.

PR #123, opened 2022-08-22
Allow to skip spec validation

cristianonicolai:
By adding a new application property, `validateSpec`, it will now be possible to bypass the Open API validation before executing the codegen.
This is specifically important for scenarios where trying to integrate with API specs that are not fully compliant.


ricardozanini:
Hi @cristianonicolai! Thanks for your constribution. Do you mind extending the description a little bit more? Thanks!

integration-tests/example-project/src/main/resources/application.properties
```diff
@@ -1,3 +1,5 @@
+quarkus.openapi-generator.codegen.validateSpec=false
```
ricardozanini, line 1: Please add another example project to the IT module. This project is already saturated with use cases.

deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/wrapper/OpenApiClientGeneratorWrapper.java
```diff
@@ -40,7 +40,8 @@ public class OpenApiClientGeneratorWrapper {
     private String apiPackage = "";
     private String modelPackage = "";
 
-    public OpenApiClientGeneratorWrapper(final Path specFilePath, final Path outputDir, final boolean verbose) {
+    public OpenApiClientGeneratorWrapper(final Path specFilePath, final Path outputDir, final boolean verbose,
```
ricardozanini, line 43: Why change the constructor's interface?

ricardozanini (commented):
I think we need to add support to any underneath property the internal generator might expose directly. Example: `quarkus.openapi-generator.codegen.<innergen>.<property_key>`. I'll create an issue.

cristianonicolai:
> se dir

@ricardozanini, that would be ideal, but unfortunately, skipping the spec validation is not available via generic properties for the openapi generator. At least in my tests, it didn't work. That's why I end up creating a new specific property.

deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/wrapper/OpenApiClientGeneratorWrapper.java
cristianonicolai: I followed the same line as it was done for setting up _verbose_. It can be done via builder too, I don't have a strong preference, either way is fine.

integration-tests/example-project/src/main/resources/application.properties
cristianonicolai: done

fjtirado (approved):

ricardozanini:
> > se dir
> 
> @ricardozanini, that would be ideal, but unfortunately, skipping the spec validation is not available via generic properties for the openapi generator. At least in my tests, it didn't work. That's why I end up creating a new specific property.

I'll keep the issue open. I think we have a few options, like using reflection since it's happening in codegen phase. Thanks for the PR!

integration-tests/pom.xml
```diff
@@ -11,6 +11,7 @@
   <packaging>pom</packaging>
   <modules>
     <module>example-project</module>
+    <module>test-project</module>
```
ricardozanini, line 14: Example-project and test-project seem ambiguous. I'll change them later to express the real intention of each use case.

ricardozanini (approved):

ricardozanini:
@all-contributors please add @cristianonicolai for code

PR #130, opened 2022-09-01
Add indexed name auth provider

cristianonicolai:


ricardozanini (commented):
LGTM, but what's the issue? Can you describe it, please?

cristianonicolai:
@ricardozanini, apparently the same provider can be added multiple times. This is the case for `oauth` because of the different flows. This way, we avoid conflict with the variable name. The attached open API spec reproduces this issue.

ricardozanini (approved):

PR #136, opened 2022-09-05
Fix #135: Exclude deprecated operations

### LINKED ISSUE #135: Option: Exclude deprecated operations from generating code
You need a Open API file with an API call marked `deprecated: true`

Here is an example operation with ` "deprecated": true,` I would love a config option `excludeDeprecatedOperations` to basically not generate any code for any deprecated APIs.  It will reduce my client code and prevent someone from calling deprecated APIs.

```json
 "/submit-re": {
      "post": {
        "tags": [
          "Manifest"
        ],
        "summary": "Deprecated please use /v2/submit-toc-file",
        "operationId": "createPackageFromFiles",
        "parameters": [
          {
            "name": "Authorization",
            "in": "header",
            "required": false,
            "schema": {
              "type": "string"
            }
          },
          {
            "name": "refresh-data",
            "in": "query",
            "required": false,
            "schema": {
              "type": "boolean",
              "default": false
            }
          }
        ],
        "requestBody": {
          "content": {
            "multipart/form-data": {
              "schema": {
                "required": [
                  "file"
                ],
                "type": "object",
                "properties": {
                  "file": {
                    "type": "string",
                    "format": "binary"
                  }
                }
              }
            }
          }
        },
        "responses": {
          "200": {
            "description": "OK",
            "content": {
              "application/json": {
                "schema": {
                  "type": "object",
                  "additionalProperties": {
                    "type": "object"
                  }
                }
              }
            }
          }
        },
        "deprecated": true,
      }

### PR DESCRIPTION
**Pull request by melloware**
OK this IF will skip generating the operation if its deprecated. 

```
 {#if !op.deprecated || openapi:genDeprecatedApiAttr(package, classname, codegen)}
```

### DIFF
--- a/deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/codegen/ModelCodegenConfigParser.java
+++ b/deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/codegen/ModelCodegenConfigParser.java
@@ -1,5 +1,6 @@
 package io.quarkiverse.openapi.generator.deployment.codegen;
 
+import static io.quarkiverse.openapi.generator.deployment.CodegenConfig.resolveApiPackage;
 import static io.quarkiverse.openapi.generator.deployment.CodegenConfig.resolveModelPackage;
 
 import java.util.HashMap;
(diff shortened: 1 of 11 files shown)

### CODE REVIEW

**Review by ricardozanini — COMMENTED** [review]
Are we deprecating the operations or the parameters within the operations? Maybe we need both, what do you think? 

Because the `<base_package>.api.MyClass.generateDeprecated=false` seems that we are skipping the operations, and indeed in your code is what you're doing.

But in the documentation, it seems like we are deprecating the attributes of the operations.

**Comment by melloware** [comment]
Good catch documentation updated to say operations and provide proper link to OpenAPI spec

**Review by ricardozanini — APPROVED** [review]

**Review by ricardozanini — COMMENTED** [review]
I forgot to add one thing, I'm sorry. Can you tweak one of the tests we have to include a deprecated API in order to create a test for this feature?

**Comment by melloware** [comment]
Yep let me do that!

**Comment by melloware** [comment]
Added unit test.  Glad I did it proved out some issues

File: deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/codegen/ModelCodegenConfigParser.java [hunk 1]
```diff
@@ -17,16 +18,19 @@
 public final class ModelCodegenConfigParser {
```
ricardozanini, line 18: Maybe we should reconsider renaming this class since now we are using it for the API generator as well. wdyt?

ricardozanini (commented):
One nitpick; it seems pretty good! Many thanks! :)

deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/codegen/ModelCodegenConfigParser.java
melloware: i agree.  on it

melloware:
Renamed "ModelCodeGen" to "ClassCodeGen"

ricardozanini (approved):

PR #137, opened 2022-09-06
Use original operationId in annotated GeneratedMethod

cristianonicolai:
The generated annotation `@GeneratedMethod` currently uses as a value a sanitized name for the `operationId`. This is necessary for codegen because it is/can be used as java identifier, for instance, in method, variable or class name.  The issue with that, is that the reference to the operationId is lost. For example, an operationId of `Jobs doJob` gets translated to `jobsDoJob`, in a method with the same name. In the case of the annotation value, it makes sense that we keep the original String, because is not used for the codegen, it is a simple string. By using the original operationId, without sanitizing, it allows consumer applications to track back with an exact match to the `operationId` contained in the OpenAPI spec file.

### DIFF
--- a/deployment/src/main/resources/templates/api.qute
+++ b/deployment/src/main/resources/templates/api.qute
@@ -65,7 +65,7 @@ public interface {classname} {
     {#if op.hasProduces}
     @Produces(\{{#for produce in op.produces}"{produce.mediaType}"{#if produce_hasNext}, {/if}{/for}\})
     {/if}
-    @GeneratedMethod ("{op.operationId}")
+    @GeneratedMethod ("{op.operationIdOriginal}")
     {#for cbClassConfig in circuit-breaker.orEmpty}{#if cbClassConfig.key == package + classname}
     {#for cbMethod in cbClassConfig.value.orEmpty}{#if cbMethod == op.nickname}
     @org.eclipse.microprofile.faulttolerance.CircuitBreaker
--- a/deployment/src/test/java/io/quarkiverse/openapi/generator/deployment/wrapper/OpenApiClientGeneratorWrapperTest.java
+++ b/deployment/src/test/java/io/quarkiverse/openapi/generator/deployment/wrapper/OpenApiClientGeneratorWrapperTest.java
@@ -166,10 +166,22 @@ public class OpenApiClientGeneratorWrapperTest {
         List<MethodDeclaration> methodDeclarations = compilationUnit.findAll(MethodDeclaration.class);
         assertThat(methodDeclarations).isNotEmpty();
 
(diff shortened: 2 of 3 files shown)


deployment/src/main/resources/templates/api.qute
```diff
@@ -65,7 +65,7 @@ public interface {classname} {
     {#if op.hasProduces}
     @Produces(\{{#for produce in op.produces}"{produce.mediaType}"{#if produce_hasNext}, {/if}{/for}\})
     {/if}
-    @GeneratedMethod ("{op.operationId}")
+    @GeneratedMethod ("{op.operationIdOriginal}")
```
spolti, line 68: is it set by openapi?

cristianonicolai: yep

ricardozanini (approved):

ricardozanini:
Maybe would make sense to have `@GeneratedMethod (operationIdOriginal= "{op.operationIdOriginal}", operationId="@GeneratedMethod ("{op.operationId}"))`

deployment/src/main/resources/templates/api.qute
spolti: I see, thanks.
What are the possible issues to use only the Original value instead the operationID?

cristianonicolai: @spolti it is the original value of the _operationId_ :) instead of the a sanitized version of it.
Without this fix, an `operationId:Jobs doJob` would result in `jobsDoJob` when we fact, we just want it to add `Jobs doJob`, which is the original value.

fjtirado (approved):

fjtirado:
> Maybe would make sense to have `@GeneratedMethod(operationIdOriginal="{op.operationIdOriginal}", operationId="{op.operationId}")`

No, I think the PR is fine, this should be the original name

ricardozanini:
> 

The only problem I see is breaking the compatibility with client code using this annotation expecting the sanitized name. That's fine, we can add a note in the release notes and we don't have a final release yet.

fjtirado:
> > 
> 
> The only problem I see is breaking the compatibility with client code using this annotation expecting the sanitized name. That's fine, we can add a note in the release notes and we don't have a final release yet.

Great @ricardozanini. One problem we have realized might occur and needs to be somehow handled by the extension is when we have two operation ids that leads to the same java method name 
This will unlikely occur, for example, if user define two operationids, one name my-method and another named my#method. 
In those cases, we should fail and force the user to define a mapping (in configuration) for these "funny" method names.

ricardozanini:
This naming clash should be handled by the OpenAPI Generator we are using. We can, however, override this behavior. I don't think this should be done here but bring it to that community instead.

PR #147, opened 2022-09-24
Introduce kotlin codegen

### LINKED ISSUE #134: Kotlin support
It would be amazing to have Kotlin support for this extension. Main advantages would be the pojos with nullability for types and coroutines for reactive implementations.

Maybe we could discuss a way to integrate it.

Happy to contribute.

stephan-strate:
Closes #134. This is a draft pr to support kotlin client codegen. Open for discussion, tips and tricks :)


deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/CodegenConfig.java
```diff
@@ -82,6 +83,10 @@ public static String getImportMappingsPropertyName(final Path openApiFilePath) {
         return String.format(IMPORT_MAPPINGS_PROP_FORMAT, getBuildTimeSpecPropertyPrefix(openApiFilePath));
     }
 
+    public static String getCodegenLanguagePropertyName(final Path openApiFilePath) {
```
stephan-strate, line 86: @ricardozanini where would you put the switch between languages?

stephan-strate:
@ricardozanini Is this the way you expected it to work? Having two sets of templates and two codegen configs

deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/CodegenConfig.java
Orbifoldt: I like it being on an individual API-spec level! (instead of it being set globally for a project) That makes the interop with java nicer and makes this useable in mixed kotlin/java projects

stephan-strate: Yes, I think this is important, because many projects are mixed!
I also thought about kotlin being the default when `quarkus-kotlin` extension is present and java being default when not. What do you think?

ricardozanini: That makes sense to me. :)

ricardozanini (commented):
Thanks, @stephan-strate! That was what I was talking about. Also, the "context-matter" approach of automatically picking kotlin as the base template when only `quarkus-kotlin` is presented in the classpath would be the cherry on top.

stephan-strate:
Just to let you know. I am still taking care of it. Need some extra time to make everything work.

Do you have some input for integration testing? I think we should reuse the existing ones and execute them with both codegen languages.

ricardozanini:
Feel free to use the existing ones. We are on the verge of refactoring the `integration-tests` modules to segregate by use case. So it will be easier to integrate and understand what's happening. cc @hbelmiro

ricardozanini:
oh, and no rush AT ALL. Do it on your time.

laurentperez:
hi. is work on this ongoing ? thx

stephan-strate:
> hi. is work on this ongoing ? thx

Currently no. Was more time consuming, than I anticipated. Started over several times, but failed to setup the test structure. My goal was to reuse existing tests and only add special tests if needed.

Still don't think, that writing the templates is much work, but the initial setup was where I got stuck.

Maybe someone else has an idea or want to give it a shot. I'm more than happy to assist.

PR #158, opened 2022-10-19
Import org.eclipse.microprofile.rest.client.annotation.RegisterProvider when custom-register-providers is used

YassinHajaj:
There is a bug actually in the released 0.11.0.
I'll still need to write / correct a few tests for this.

### DIFF
--- a/deployment/src/main/resources/templates/api.qute
+++ b/deployment/src/main/resources/templates/api.qute
@@ -5,8 +5,10 @@ import {imp.import};
 {/for}
 
 import org.eclipse.microprofile.rest.client.inject.RegisterRestClient;
+{#if hasAuthMethods || custom-register-providers.orEmpty.size > 0}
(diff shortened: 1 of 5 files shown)


ricardozanini:
Thank you @YassinHajaj. Now your comment on the other issue makes more sense. Indeed is a case that we didn't anticipate.

YassinHajaj:
Hey @ricardozanini ,
You're welcome :) I think the PR is ready now.
By the way, do you have any idea when we can expect the 0.12.0 ?
If there's anything else I can help with, don't hesitate

hbelmiro:
> I believe we can release it after this PR is merged. cc @hbelmiro wdyt?

I agree @ricardozanini.

hbelmiro (approved):
LGTM.

ricardozanini:
@all-contributors please add @YassinHajaj for code

Issue #39, opened 2022-04-01
Allow to exclude/incldue some definition from code generation


lburgazzoli:
In [this](https://github.com/lburgazzoli/bf2-cos-fleetshard/tree/opeanpi-get-quarus/cos-fleet-manager-api/src/main/openapi) repo I have a number of openapi files: 
- `connector_mgmt-private.yaml`
- `connector_mgmt.yaml`
- `kas-fleet-manager.yaml`

In this set up, `connector_mgmt-private.yaml` is the main entry point for code generation and the others, are only referenced by `connector_mgmt-private.yaml`. It looks like it is not possible to configure the extension to include/exclude some of the definitions file present in the openapi folder for code generation.

PR #171, opened 2022-11-04
[issue-39] Added feature to ignore code generation for specific OpenA…

hbelmiro:
…PI documents

Closes: https://github.com/quarkiverse/quarkus-openapi-generator/issues/39

### DIFF
--- a/deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/codegen/OpenApiGeneratorCodeGenBase.java
+++ b/deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/codegen/OpenApiGeneratorCodeGenBase.java
@@ -26,7 +26,6 @@ import static io.quarkiverse.openapi.generator.deployment.CodegenConfig.getImpor
 import static io.quarkiverse.openapi.generator.deployment.CodegenConfig.getSanitizedFileName;
 import static io.quarkiverse.openapi.generator.deployment.CodegenConfig.getSkipFormModelPropertyName;
 import static io.quarkiverse.openapi.generator.deployment.CodegenConfig.getTypeMappingsPropertyName;
-import static java.util.function.Predicate.not;
 
 /**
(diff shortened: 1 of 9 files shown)


deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/codegen/OpenApiGeneratorCodeGenBase.java
```diff
@@ -41,11 +51,14 @@ public String inputDirectory() {
...
         final Path outDir = context.outDir();
         final Path openApiDir = context.inputDir();
+        final List<String> ignoredFiles = context.config()
+                .getOptionalValues("quarkus.openapi-generator.codegen.ignore", String.class).orElse(List.of());
 
         if (Files.isDirectory(openApiDir)) {
             try (Stream<Path> openApiFilesPaths = Files.walk(openApiDir)) {
                 openApiFilesPaths
                         .filter(Files::isRegularFile)
+                        .filter(not(path -> ignoredFiles.contains(path.getFileName().toString())))
```
fjtirado, line 61: Should now we move this filter after the Map::toString
I have to confess there is to many streaming here for my taste ;), but it we are gonna play streaming, then lets play it smartly, below this filter is a string conversion and another filter, so lets put it there
I would also argue that this filter should be the last one of the chain, because it is the slower one

hbelmiro: No, because `Path#toString` returns the full path and I need only the filename. Note that I use `path.getFileName().toString()` in the second filter.

hbelmiro: Done.

ricardozanini (commented):
Hi! Thanks! The implementation seems fine, but what's the motivation for this feature? Do we have a use case for it? Can you please describe it in the PR description?

ricardozanini (approved):

PR #189, opened 2022-11-30
Added info about Quarkus 3 to README.md

hbelmiro:
Fixes: https://github.com/quarkiverse/quarkus-openapi-generator/issues/191

PR to support Quarkus 3 in `main` branch: https://github.com/quarkiverse/quarkus-openapi-generator/pull/190

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,6 +10,8 @@
 
 > **⚠️** This is the instructions for the latest SNAPSHOT version (main branch). Please, see the [latest **released** documentation](https://github.com/quarkiverse/quarkus-openapi-generator/blob/0.12.0/README.md) if you are looking for instructions.
 
+> **⚠️** Version 2.x.x of this extension (`main` branch) supports Quarkus 3, and version 1.x.x (`quarkus2` branch) supports Quarkus 2.
+
 Quarkus' extension for generation of [Rest Clients](https://quarkus.io/guides/rest-client) based on OpenAPI specification files.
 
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -8,7 +8,7 @@
   </parent>
   <groupId>io.quarkiverse.openapi.generator</groupId>
   <artifactId>quarkus-openapi-generator-parent</artifactId>
-  <version>1.0.0-SNAPSHOT</version>
+  <version>1.0.0-quarkus2-SNAPSHOT</version>
```
gsmet, line 11: I think we would rather go with a major version bump for the version compatible with Quarkus 3.

@gastaldi WDYT?

hbelmiro: @ricardozanini ^

ricardozanini: Agreed. I'll release the last 0.x version today, and for Quarkus 3 support, we do 1.x.

hbelmiro: @ricardozanini I think they are suggesting 0.x to Quarkus 2 and 1.x to Quarkus 3 and both keep evolving rather than having 1.x.x for Quarkus 3 and 1.x.x-quarkus2 for Quarkus 2.

@gsmet and @gastaldi, can you clarify if my understanding is correct?

ricardozanini: That's what I meant 😆

ricardozanini (approved):

Issue #196, opened 2022-12-05
Unexpected exception when sourceDir doesn't include "src"


gwydionmv:
The change for configuring the inputBaseDir property includes the method `OpenApiGeneratorCodeGenBase.getInputBaseDirRelativeToModule`. It generates the directory relative to the module by substringing the original path according to the 'src' string. But if the original path doesn't include 'src', it's failing uncontrolledly. Opening issue to control this exception and show a clearer message.

PR #197, opened 2022-12-05
Control when path has not 'src' in it

gwydionmv:
Fixing issue #196 

An exception was being thrown when trying to substring a path not including 'src' in it. Now it's controlled that the index for the substring is valid before doing it.

### DIFF
--- a/deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/codegen/OpenApiGeneratorCodeGenBase.java
+++ b/deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/codegen/OpenApiGeneratorCodeGenBase.java
@@ -140,7 +140,9 @@ public abstract class OpenApiGeneratorCodeGenBase implements CodeGenProvider {
     }
 
     private String getInputBaseDirRelativeToModule(final Path sourceDir, final Config config) {
-        String baseModuleDirectory = sourceDir.toString().substring(0, sourceDir.toString().lastIndexOf("src"));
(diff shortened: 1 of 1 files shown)


hbelmiro (commented):
@gwydionmv do you mind including a test for this use case?

gwydionmv:
> I'm not exactly sure how I can manually force a Path for the execution of the plugin. I faced this issue because another plugin we are using in the project is setting it up internally

hbelmiro:
@gwydionmv can't you just define a path without src with the property?

quarkus.openapi-generator.codegen.input-base-dir=openapi-definitions

gwydionmv:
@hbelmiro the property it is relative to the project base directory. E.g. if you have a project called `MyProject` and decide to place the openapi file in `MyProject/openapi-definitions`, the property as you wrote it is valid.

The lack of src in the path is not related to the property, but to the sourceDir that's being set up automatically by maven and the quarkus compiler plugin. It's something similar to the issue we opened last week for the /test directory: it sets different paths depending on I don't know what yet. For your example tests, it adds the /test; for some of our projects, only /src/main; for others, it also includes /target/generated-sources...

To sum up, it's not something we can configure via input-base-dir property.

hbelmiro (approved):
LGTM.
Thanks for the contribution @gwydionmv.

gwydionmv:
@ricardozanini please check this out when you have a moment to spare :)

ricardozanini (approved):
I'll release a patch version to include this fix. Thank you, @gwydionmv

PR #217, opened 2023-01-02
Prepare Release for 1.1.0

ricardozanini:


.github/project.yml
```diff
@@ -1,3 +1,3 @@
 release:
-  current-version: 0.12.0
-  next-version: 1.0.0-SNAPSHOT
+  current-version: 1.1.0
+  next-version: 2.0.0-SNAPSHOT
```
hbelmiro, line 3: We need to increment this also on main branch.

hbelmiro (approved):

.github/project.yml
hbelmiro: I mean, on main:
next-version: 3.0.0-SNAPSHOT

ricardozanini: I know, this is just here.

Issue #174, opened 2022-11-14
Make sure that the property `interfaceOnly` is exposed


ricardozanini:
See: https://openapi-generator.tech/docs/generators/jaxrs-spec/

We already use this generator underneath. The property `interfaceOnly` can be exposed by the plugin. The trick here is also to update the templates to ensure that the property is considered.

Request: https://github.com/quarkiverse/quarkus-openapi-generator/issues/128#issuecomment-1310557441

KiMurray:
Im a little confused. 
> We already use this generator underneath.
My understanding was that this extension uses the Java client generator, has something changed? / Can this extension also generate server stubs? 
Cheers

ricardozanini:
We use the Java Client Generator from the OpenAPI Generator library. Theoretically we can generate server stubs too, but it's out of scope. 

The idea behind this issue is to implement a similar behavior in the client side, so instead of generating something like this:

```java
@Produces(MediaType.APPLICATION_JSON)
@Path("/petstore")
public class PetResource {

    @RestClient
    @Inject
    PetApi petApi;
}
```

We do this:

```java
@Produces(MediaType.APPLICATION_JSON)
@Path("/petstore")
public class PetResource {

    @RestClient
    @Inject
    Response petApi;
}
```

@KiMurray that's what you're looking for, right?

hbelmiro:
Sorry @KiMurray. It seems to be the same as @ricardozanini is saying in his previous comment.
Am I missing something?

KiMurray:
@ricardozanini, yes! your updated comment is exactly what we need. Had me a little confused for a moment there. 😄

hbelmiro:
@ricardozanini this issue's requirement is not very clear.
My understanding from @KiMurray's [original comment](https://github.com/quarkiverse/quarkus-openapi-generator/issues/128#issuecomment-1310557441) is 2 requirements:
1) Generate server code
This could be achieved by exposing `interfaceOnly`, but this is out of the scope of this extension and we won't implement it.

2) Return `Response` instead of a model
Is this related with `interfaceOnly`? If yes, I think we shouldn't use `interfaceOnly` to enable/disable this, since it may confuse users and actually wouldn't describe the "return `Response`" feature.

ricardozanini:
`interfaceOnly` is just a name. As soon as we start working on the enhancement, we can develop a new name for this property.

[...]

PR #219, opened 2023-01-05
Added "return-response" property to return `Response` instead of a model

hbelmiro:
Fixes: https://github.com/quarkiverse/quarkus-openapi-generator/issues/174

I named the property `return-response` since `interfaceOnly` doesn't describe this feature and might confuse users. I'm also open to better property names.

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,11 +121,11 @@ See the [integration-tests](integration-tests) module for more information of ho
 
 ## Returning `Response` objects
 
-By default, this extension generates methods that return `void`, values as `String`, `Integer`, etc., or an object model, like `Product` for instance. If you always want to return `jakarta.ws.rs.core.Response`, you can set the `return-response` property.
+By default, this extension generates the methods according to their returning models based on the [OpenAPI specification Schema Object](https://spec.openapis.org/oas/v3.1.0#schema-object). If you want to return `jakarta.ws.rs.core.Response` instead, you can set the `return-response` property to `true`.
 
 ### Example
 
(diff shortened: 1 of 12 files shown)


README.md
```diff
@@ -119,6 +119,18 @@ public class PetResource {
 
 See the [integration-tests](integration-tests) module for more information of how to use this extension. Please be advised that the extension is on experimental, early development stage.
 
+## Returning `Response` objects
+
+By default, this extension generates methods that return `void`, values as `String`, `Integer`, etc., or an object model, like `Product` for instance. If you always want to return `jakarta.ws.rs.core.Response`, you can set the `return-response` property.
+
```
ricardozanini, line 125: You can link here the properties from the OpenAPI describing that we generate the model based on that.

ricardozanini (commented):
I don't know if I like this name, but I can't find a better one now. :(

README.md
hbelmiro: I'm not sure it would be useful. The docs don't mention this behavior.

image

https://openapi-generator.tech/docs/generators/jaxrs-spec/

ricardozanini: No, I mean this section: https://spec.openapis.org/oas/v3.1.0#schema-object

Linking here and say something like, "Usually, we generate the methods according to their returning models based on the Open API spec schema object. Turn this property to `true` to generate JAXRS `Response` instead.

hbelmiro: Done.

ricardozanini (approved):

hbelmiro:
Cherry-picked to quarkus2 branch: https://github.com/quarkiverse/quarkus-openapi-generator/pull/223

Issue #220, opened 2023-01-05
Can't start project without `base-package` property set


hbelmiro:
If we don't set the `base-package` property for any OpenAPI document, the application can't start. It gives the following error:

```bash
2023-01-05 10:30:33,774 ERROR [io.qua.dep.dev.IsolatedDevModeMain] (main) Failed to start quarkus: io.quarkus.runtime.configuration.ConfigurationException: SRCFG00014: The config property quarkus.openapi-generator.codegen.spec.openapi_yaml.base-package is required but it could not be found in any config source
	at io.quarkus.deployment.configuration.BuildTimeConfigurationReader$ReadOperation.readConfigValue(BuildTimeConfigurationReader.java:784)
	at io.quarkus.deployment.configuration.BuildTimeConfigurationReader$ReadOperation.readConfigGroup(BuildTimeConfigurationReader.java:722)
	at io.quarkus.deployment.configuration.BuildTimeConfigurationReader$ReadOperation.recreateGroup(BuildTimeConfigurationReader.java:679)
	at io.quarkus.deployment.configuration.BuildTimeConfigurationReader$ReadOperation.getGroup(BuildTimeConfigurationReader.java:612)
	at io.quarkus.deployment.configuration.BuildTimeConfigurationReader$ReadOperation.run(BuildTimeConfigurationReader.java:389)
	at io.quarkus.deployment.configuration.BuildTimeConfigurationReader.lambda$readConfiguration$0(BuildTimeConfigurationReader.java:322)
	at io.smallrye.config.SecretKeys.doUnlocked(SecretKeys.java:29)
	at io.quarkus.deployment.configuration.BuildTimeConfigurationReader.readConfiguration(BuildTimeConfigurationReader.java:322)
	at io.quarkus.deployment.ExtensionLoader.loadStepsFrom(ExtensionLoader.java:195)
	at io.quarkus.deployment.QuarkusAugmentor.run(QuarkusAugmentor.java:107)
	at io.quarkus.runner.bootstrap.AugmentActionImpl.runAugment(AugmentActionImpl.java:331)
	at io.quarkus.runner.bootstrap.AugmentActionImpl.createInitialRuntimeApplication(AugmentActionImpl.java:252)
	at io.quarkus.runner.bootstrap.AugmentActionImpl.createInitialRuntimeApplication(AugmentActionImpl.java:60)
[...]

PR #221, opened 2023-01-05
Fixed issue that prevented the app from starting when "base-package" property was not set

hbelmiro:
Fixes: https://github.com/quarkiverse/quarkus-openapi-generator/issues/220


ricardozanini (approved):

ricardozanini:
Then what, the application is generated using the default prefix only?

hbelmiro:
The application was generated correctly when the property was not present, but Quarkus didn't allow it to start. This change doesn't change the behavior. It just informs Quarkus that `base-package` is an optional property.

hbelmiro:
Cherry-picked to quarkus2 branch: https://github.com/quarkiverse/quarkus-openapi-generator/pull/222

Issue #150, opened 2022-10-04
Migrate to RESTEasy Reactive


hbelmiro:
This extension uses RESTEasy Classic. However, Quarkus recommends using RESTEasy Reactive now.
Documentation: https://quarkus.io/guides/resteasy-reactive-migration

melloware:
I would like a switch bewteen the two if possible because a lot of us still use RestEasy Classic.

melloware:
Isn't this also similar to this ticket: https://github.com/quarkiverse/quarkus-openapi-generator/issues/82

Read that discussion where it talks about generating different client types.

hbelmiro:
@melloware I don't think so. This is just to replace the RESTEasy classic dependency with the reactive one. It shouldn't result in significant changes.

hbelmiro:
There's this compatibility break that affects Multipart requests: https://github.com/quarkusio/quarkus/issues/23059#issue-1109278469
We need a definition from Quarkus before proceeding with this.

melloware:
Good I am heavily using MultiPart requests.

ricardozanini:
Relates to #157

einarjohnson:
Any ETA on when this might be fixed/become compatible?

hbelmiro:
> Any ETA on when this might be fixed/become compatible?

@einarjohnson I'm working on this.
There's [this compatibility](https://github.com/quarkusio/quarkus/issues/23059#issuecomment-1287053067) break I still need to assess and there might have others that I'm still not aware of. Therefore I don't have a ETA. Sorry.

einarjohnson:
@hbelmiro, thanks for getting back to me so quickly.

ssaip:
Very much needed feature! Until this is done, we have to stick to the normal openapi-generator :(

ricardozanini:
@ssaip unfortunately, the compatibility break, as documented in https://github.com/quarkusio/quarkus/issues/23059#issuecomment-1287053067 prevents us from migrating.

We can migrate, but we will lose a feature. We can hear from the community and people following this thread whether or not the benefit of using RestEasy reactive is ok to lose the feature.

@hbelmiro, or we can study a solution for users to pick their favorite. We will have to keep a few different templates, but worth having if someone is interested in contributing.

hbelmiro:
[...]

PR #262, opened 2023-02-20
Add support to RESTEasy Reactive

hbelmiro:
Fixes: https://github.com/quarkiverse/quarkus-openapi-generator/issues/150


deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/codegen/RestEasyImplementationVerifier.java
```diff
@@ -0,0 +1,49 @@
+package io.quarkiverse.openapi.generator.deployment.codegen;
+
+import java.util.Collection;
+
+final class RestEasyImplementationVerifier {
+
+    private static final RestEasyImplementationVerifier INSTANCE = new RestEasyImplementationVerifier();
+
+    private static final String BOTH_DEPENDENCIES_PRESENT_ERROR_MSG = "The application has RESTEasy classic and reactive on its dependencies. Both dependencies cannot coexists. You need to remove one of them.";
```
ricardozanini, line 9: I thought they already have this check in their codebase.

integration-tests/change-directory/src/main/resources/application.properties
```diff
@@ -1,2 +1,3 @@
 quarkus.openapi-generator.codegen.input-base-dir=openapi-definitions
 quarkus.openapi-generator.codegen.spec.simple_openapi_yaml.base-package=org.simple.openapi
+quarkus.rest-client.simple_openapi_yaml.uri=http://localhost:8080
```
ricardozanini, line 3: Is this required now?

integration-tests/multipart-request/src/main/openapi/multipart-requests.yml
```diff
@@ -62,7 +62,7 @@ components:
           example: New Delhi
     UserId:
       type: string
-      format: uuid
+      format: string
```
ricardozanini, line 65: What's the problem with this format?

deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/wrapper/OpenApiClientGeneratorWrapper.java
```diff
@@ -151,6 +151,11 @@ public OpenApiClientGeneratorWrapper withAdditionalModelTypeAnnotationsConfig(fi
         return this;
     }
 
+    public OpenApiClientGeneratorWrapper withRestEasyReactive(Boolean isRestEasyReactive) {
```
ricardozanini, line 154: Not sure if this is a good idea. We should resolve it internally, like you're doing on `OpenApiGeneratorCodeGenBase`, not expose it in the wrapper.

ricardozanini (commented):
Looks neat! Many thanks. I was debating rather have specific templates for classic/reactive, but I think the changes have less impact on them then I initially thought. Just a few comments, if you don't mind. 

Also, it's worth taking a look at #263 to see if there's anything we might miss here. I haven't looked at it thou.

integration-tests/change-directory/src/main/resources/application.properties
hbelmiro: Yes.

integration-tests/multipart-request/src/main/openapi/multipart-requests.yml
hbelmiro: Quarkus still doesn't support it. It will be supported in the next version (2.16.4).
I can add a converter if you prefer, as I did for OffsetDateTime in https://github.com/quarkiverse/quarkus-openapi-generator/pull/262/files#diff-61172ef4aaa50da592ec53387fd6765f66e2e9c8f58feba1539a7a742112b447.

https://github.com/quarkusio/quarkus/issues/31298

hbelmiro: The quarkus issue: https://github.com/quarkusio/quarkus/issues/31298

ricardozanini: Thank u!

integration-tests/type-mapping/src/main/java/io/quarkiverse/openapi/generator/it/type/mapping/OffsetDateTimeParamConverter.java
```diff
@@ -0,0 +1,18 @@
...
+    @Override
+    public OffsetDateTime fromString(String value) {
+        return value == null ? null : OffsetDateTime.parse(value);
+    }
+
+    @Override
+    public String toString(OffsetDateTime value) {
+        return value == null ? null : value.toString();
+    }
+}
```
chberger, line 18: Looks like my code :-)

runtime/src/main/java/io/quarkiverse/openapi/generator/providers/OAuth2AuthenticationProvider.java
```diff
@@ -57,7 +57,7 @@ private void validateConfig() {
         }
     }
 
-    public interface OidcClientRequestFilterDelegate {
-        void filter(ClientRequestContext requestContext) throws IOException;
+    public interface OidcClientRequestFilterDelegate<T extends ClientRequestContext> {
```
chberger, line 60: Cool, even better than my solution!

runtime/pom.xml
```diff
@@ -15,10 +15,6 @@
       <groupId>io.quarkus</groupId>
       <artifactId>quarkus-core</artifactId>
     </dependency>
-    <dependency>
-      <groupId>io.quarkus</groupId>
-      <artifactId>quarkus-rest-client-jackson</artifactId>
-    </dependency>
     <dependency>
       <groupId>io.quarkus</groupId>
       <artifactId>quarkus-qute</artifactId>
```
chberger, line 20: I would recommend to set the scope of ``quarkus-oidc-client-filter`` from ``provided`` to ``test``.  The dependency is required by test classes only and at runtime the app might define a ``reactive oidc filter`` dependency. And a comment explaining why a classic dependency is required would help as well (see my PR)

chberger (commented):
Overall, the implementation is very similar to my one. However, using the ``CodeGenContext`` to derive the ``rest-client mode`` is much better than my feature flag. I've added some comments which might be incorporated. In addition to that, the ``QuteTemplatingEngineAdapterTest`` throws an exception on Windows which is ignored by the test assertion. I've fixed that in my PR. Furthermore, I've spent some time updating the ``README`` with all the new implications (multipart changes etc). Would love to see that in your PR as well ... Thanks!

deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/wrapper/OpenApiClientGeneratorWrapper.java
hbelmiro: Done.

deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/codegen/OpenApiGeneratorCodeGenBase.java
```diff
@@ -70,6 +72,14 @@ public boolean shouldRun(Path sourceDir, Config config) {
...
     }
 
+    protected boolean isRestEasyReactive(CodeGenContext context) {
+        List<String> appDependencies = context.applicationModel().getDependencies().stream()
+                .map(ResolvedDependency::getArtifactId)
+                .collect(Collectors.toList());
+
+        return RestEasyImplementationVerifier.get().isRestEasyReactive(appDependencies);
+    }
+
```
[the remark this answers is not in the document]
hbelmiro: I changed it to use `ExtensionCapabilities`. But anyway, I'm not sure if it makes any difference to your question. Why couldn't it work with Gradle?

deployment/src/main/resources/templates/api.qute
```diff
@@ -80,11 +80,19 @@ public interface {classname} {
     public {#if op.returnType}{op.returnType}{#else}void{/if} {op.nickname}(
         {/if}
         {#if op.hasFormParams}
+        {#if is-resteasy-reactive}
+        @javax.ws.rs.BeanParam {op.operationIdCamelCase}MultipartForm multipartForm{#if op.hasPathParams},{/if}{!
```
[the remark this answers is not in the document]
hbelmiro: You mean the org.jboss.resteasy.reactive.MultipartForm annotation? I'm not using it.

integration-tests/array-enum/pom.xml
```diff
@@ -46,6 +46,27 @@
...
+    </profile>
+    <profile>
+      <id>resteasy-reactive</id>
+      <dependencies>
+        <dependency>
+          <groupId>io.quarkus</groupId>
+          <artifactId>quarkus-rest-client-reactive-jackson</artifactId>
+        </dependency>
+      </dependencies>
+    </profile>
```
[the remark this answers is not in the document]
hbelmiro: I declared the profiles in sub-modules because each integration test has its own dependencies. If we declare the profiles in the `integration-tests` module, we'll need to declare the dependencies used by all tests.
But thinking better now, this might be a good price to pay in order to make it simpler. WDYT @ricardozanini ?

runtime/pom.xml
hbelmiro: Actually, there are several non-test classes that use it. And that's a problem. I think I'll need to create a reactive and a classic version for each of these classes.

deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/codegen/OpenApiGeneratorCodeGenBase.java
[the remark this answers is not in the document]
chberger: I'm just wondering why ``ResolvedDependency `` comes from ``io.quarkus.**maven**.dependency`` and would that work if an app would use ``Gradle`` to resolve the ``rest-client-reactive`` Quarkus extension ... 

Anyway, since you moved to``ExtensionCapabilities`` you'll check capabilities instead of resolved artifacts which should work in any case ..

deployment/src/main/resources/templates/api.qute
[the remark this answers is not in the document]
chberger: You're right, my fault!

runtime/pom.xml
hbelmiro: Good catch!
Done. I had to add a couple of dependencies that we lose when we reduce the scope of `quarkus-oidc-client-filter`.

deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/codegen/OpenApiGeneratorCodeGenBase.java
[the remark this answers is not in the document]
hbelmiro: Got it ;)
Thanks!

integration-tests/array-enum/pom.xml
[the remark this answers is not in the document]
hbelmiro: Done.

chberger (commented):
Thanks @hbelmiro for your contribution. Overall, the PR looks good to me. I've also successfully tested your version with my current project which uses ``rest-client-reactive``. 

I would still have a few small details: Since the templates rely on ``Jackson`` and due to the fact that we removed ``quarkus-rest-client-jackson`` from the runtime module (for good reason), there is no guarantee that Jackson is available at runtime. Thus I would recommend a ``BuildStep`` which checks (yes during startup) the required prerequisites for convenience purposes only. I know it would fail at runtime anyway but I guess this would improve the user experience especially when we print some concrete tasks like "please add ``quarkus-rest-client-jackson`` dependency e.g. in case ``rest-client-reactive`` has been detected. Furthermore it would like to see an updated ``README.md``, especially since ``reactive`` support is available now.

Beside that there is still this problem with the ``InvalidPathException`` at QuteTemplatingEngineAdapterTest.java:27 on Windows, which is ignored by the assertion. I've fixed that in my PR --> https://github.com/quarkiverse/quarkus-openapi-generator/pull/263/files#diff-2b8bfdfff562023eb3106df65d5a93dbe9cf16c80a33ef91a745a13b2eadcb58. But since this isn't actually related to the ``reactive``-activity, I could open a new issue if you want to..

hbelmiro:
@chberger thank you for reviewing.
Those are good points. I'll add more commits to the PR covering them.

Regarding the ignored assertion, please go ahead and open the issue.

hbelmiro:
@chberger @ricardozanini I've added validation to check if Jackson is in the dependencies. However, I had to add it in compile time because the execution doesn't even get in build time (the lack of the Jackson dependency causes an error in compile time). So, build-time is too late.

Also, I had to check for the reactive dependency without using `ExtensionCapabilities` because Quarkus doesn't provide it.

README.md
```diff
@@ -122,6 +122,12 @@ public class PetResource {
 
 See the [integration-tests](integration-tests) module for more information of how to use this extension. Please be advised that the extension is on experimental, early development stage.
 
+## RESTEasy Reactive and Classic support
+
+This extension supports both RESTEasy implementations (Classic and Reactive). The generated code is based on the implementation that is present in the application's dependencies.  
+
+For both implementations, the generated code is always blocking code.
+
```
chberger, line 130: Thanks, but I guess more changes are required.

In section ``OAuth2 Authentication`` the Quarkus OIDC Client Filter Extension for ``classic`` has been mentioned only. Obviously, for ``reactive`` we need another one.

In section ``Sending multipart/form-data`` it's described that the dependency ``quarkus-reasteasy-multipart`` is required, which is not true for the ``reactive`` use case. The reactive client supports ``multipart`` out of the box. In addition, the provided examples are just valid for ``classic``, not ``reactive``.

In section ``Known Limitations`` the ``no reactive support`` needs to be removed. 

I hope I got everything

deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/codegen/OpenApiGeneratorCodeGenBase.java
```diff
@@ -98,20 +119,34 @@ public boolean trigger(CodeGenContext context) throws CodeGenException {
         return false;
     }
 
+    private boolean isJacksonReactiveClientPresent(CodeGenContext context) {
+        return context.applicationModel().getDependencies().stream()
+                .anyMatch(this::isJacksonReactiveClient);
+    }
```
chberger, line 125: Damn, you're right. ``quarkus-rest-client-reactive-jackson`` doesn't define capabilities, whereas ``quarkus-rest-client-jackson`` does. I can well imagine that they just missed that, because it's just a definition in ``quarkus-extension.yaml`` like:
```
  capabilities:
    provides:
    - "io.quarkus.rest.client.reactive.jackson"
```
Maybe it make sense to raise an issue at the Quarkus project. I can do that if you like. 

However, to be backward compatible we should go with analyzing the dependencies, although I'm still wondering why the ``ResolvedDependency`` class comes from a ``Quarkus Maven`` package and whether your code would also work in case a business app would use ``Gradle``.

chberger (commented):
@hbelmiro Please look at my comments again, thanks!

README.md
hbelmiro: @chberger I've copied the changes you pointed out from your PR. Now I think it's all there.

@ricardozanini the features are the same for reactive and classic. They only have slight differences, which are now pointed out.

deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/codegen/OpenApiGeneratorCodeGenBase.java
hbelmiro: Yes. Please open an issue if you don't mind.
You can also ask about this Maven/Gradle in the issue and link this conversation to contextualize them. In case this is not the best approach, they can suggest a better one.

chberger (approved):
LGTM - thanks!

deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/codegen/OpenApiGeneratorCodeGenBase.java
chberger: @hbelmiro FYI https://github.com/quarkusio/quarkus/issues/31756#issuecomment-1463651307

ricardozanini (approved):

hbelmiro:
Cherry-picked to `main` branch: https://github.com/quarkiverse/quarkus-openapi-generator/pull/274

PR #263, opened 2023-02-21
feat: REST Client Reactive support added

chberger:
This PR contains my approach to add `REST Client Reactive` support to the `quarkus-openapi-generator`. However, since the changes have some implications I'll mention the most relevant ones below:

- I've removed all `REST Client Classic` dependencies from the deployment / runtime module. These dependencies have to be defined by the application in the future. This is the only breaking change, cause `quarkus-rest-client-jackson` cannot be resolved transitively anymore. An application needs to define the dependency explicitly. Apart from that, everything should work for `REST Client Classic` as before.
- I've adapted the integration tests slightly to make them pass with `REST Client Classic` and `REST Client Reactive` respectively.
- When executing integration tests, by default a profile called `rest-client-classic` is activated which pulls in the required `classic` dependencies. In contrast to that, you can active a `rest-client-reactive` profile to execute the same test-suite with `REST Client Reactive` dependencies enabled.
- I've spent a lot of time to come up with a solution that determines whether `REST Client Reactive` or `REST Client Classic` shall be used. The problem is that we need that info at the `code-gen` phase already. That's way earlier than any BuildStep execution. Thus I couldn't use the `Capabilities` as suggested. So I've spent some time with the TCCL to find reactive classes on the classpath, but they are somehow not present, or only in specific circumstances. Very strange behavior ... Unfortunately, this is my first touch point with Quarkus extensions so I haven't insights about Quarkus class-loading yet. Anyway, at the end I've implemented a feature toggle to define which `mode` should be used: `quarkus.openapi-generator.codegen.rest-client-reactive=true`. This property is codegen phase fixed (before build time). At least the configuration is verified at runtime and in case of misconfiguration some hints are printed.
- When going with `reactive` some features behave differently:
  -  E.g. multiple MIME-TYPES at `@Consumes` aren't supported. Seems like this is an issue with the MP REST Client Spec and the impl behavior is undefined https://github.com/eclipse/microprofile-rest-client/issues/110
  - In case of `multipart/form-data` I've kept the `MultipartForm` annotation to be backward compatible, although it's deprecated. I guess this should be changed in the future.
  - Some complex data types might need to be mapped by the user on its one. Please see `ParamConverter` for more details.
  - Not sure whether the `OIDC` filter stuff really works. :-)

The integration tests can be triggered for reactive with the following command:
```
mvn clean test -Dquarkus.openapi-generator.codegen.rest-client-reactive=true -Prest-client-reactive
```

Any feedback is appreciated.


chberger:
won't be merged in favor of #262

Issue #267, opened 2023-03-08
InvalidPathException: getResource puts a leading slash before the disk name on Windows


chberger:
When executing the test method ``checkTemplateGenerator`` in ``QuteTemplatingEngineAdapterTest`` an exception is thrown on Windows cause the swagger definition can't be loaded:

```
INFO] Running io.quarkiverse.openapi.generator.deployment.template.QuteTemplatingEngineAdapterTest
2023-03-08 16:06:38,527 WARN  [io.swa.v3.par.OpenAPIV3Parser] (main) Exception while reading:: io.swagger.v3.parser.exception.ReadContentException: Unable to read location `/C:/Users/*******/quarkus-openapi-generator/deployment/target/test-classes/openapi/petstore-openapi.json`
        at io.swagger.v3.parser.OpenAPIV3Parser.readContentFromLocation(OpenAPIV3Parser.java:238)                                                                                                                                                                                                        
        at io.swagger.v3.parser.OpenAPIV3Parser.readLocation(OpenAPIV3Parser.java:87)                                                                                                                                                                                                                    
        at io.swagger.parser.OpenAPIParser.readLocation(OpenAPIParser.java:16)                                                                                                                                                                                                                           
        at org.openapitools.codegen.config.CodegenConfigurator.toContext(CodegenConfigurator.java:525)                                                                                                                                                                                                   
        at org.openapitools.codegen.config.CodegenConfigurator.toClientOptInput(CodegenConfigurator.java:583)                                                                                                                                                                                            
        at io.quarkiverse.openapi.generator.deployment.template.QuteTemplatingEngineAdapterTest.checkTemplateGenerator(QuteTemplatingEngineAdapterTest.java:27)                                                                                                                                          
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)                                                                                                                                                                                                                
        at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)                                                                                                                                                                                              
        at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)                                                                                                                                                                                      
        at java.base/java.lang.reflect.Method.invoke(Method.java:568)                                                                                                                                                                                                                                    
[...]

PR #268, opened 2023-03-08
fixes: getResource puts a leading slash before the disk name on Windows

chberger:
@hbelmiro please find attached my PR for issue https://github.com/quarkiverse/quarkus-openapi-generator/issues/267

### DIFF
--- a/deployment/src/test/java/io/quarkiverse/openapi/generator/deployment/template/QuteTemplatingEngineAdapterTest.java
+++ b/deployment/src/test/java/io/quarkiverse/openapi/generator/deployment/template/QuteTemplatingEngineAdapterTest.java
@@ -19,11 +19,13 @@ public class QuteTemplatingEngineAdapterTest {
     @Test
     void checkTemplateGenerator() throws IOException {
         final String petstoreOpenApi = requireNonNull(this.getClass().getResource("/openapi/petstore-openapi.json")).getPath();
+        // getResource puts a leading slash before the disk name on Windows, thus we wrap it into a file object
+        final File petstoreOpenApiFile = new File(petstoreOpenApi);
         final DefaultGenerator generator = new DefaultGenerator();
         final CodegenConfigurator configurator = new QuarkusCodegenConfigurator();
-        final File apiFile = File.createTempFile("api", "java");
+        final File apiFile = File.createTempFile("api", ".java");
         apiFile.deleteOnExit();
-        configurator.setInputSpec(petstoreOpenApi);
+        configurator.setInputSpec(petstoreOpenApiFile.getAbsolutePath());
         generator.opts(configurator.toClientOptInput());
 
         final File writtenFile = generator.getTemplateProcessor().write(Collections.singletonMap("name", "Jack"), "hello.qute",


ricardozanini:
@chberger is this error only on the test case or it's a problem in general with the module? I think we need not to swallow this exception.

chberger:
@ricardozanini This is the behavior of ``io.swagger.v3.parser.OpenAPIV3Parser``, so nothing we can fix. However, they return a ``SwaggerParseResult´´ containing some error data.

hbelmiro (approved):
Many thanks for this contribution, @chberger.
LGTM

ricardozanini (approved):

hbelmiro:
Cherry-picked to main branch: https://github.com/quarkiverse/quarkus-openapi-generator/pull/269

hbelmiro:
> > @ricardozanini This is the behavior of `io.swagger.v3.parser.OpenAPIV3Parser`, so nothing we can fix. However, they return a ``SwaggerParseResult´´ containing some error data.
> 
> 
> 
> Thank you! Well, I think we should throw an exception if `SwaggerParseResult` contains error data. Not in this PR, thou. @hbelmiro wdyt? We should halt in Compile/Build time since it can be too late to find an error in the openapi definition.

@ricardozanini I agree with you.

ricardozanini:
Can you open an issue to track this?

ricardozanini:
@allcontributors please add @chberger for code

chberger:
> Can you open an issue to track this?

@ricardozanini I've analyzed the issue once more. It's something which needs to be fixed upstream (io.swagger.parser.v3:swagger-parser). Nothing we can do ... Even the ``SwaggerParseResult`` containing the error gets swallowed and we have no chance to react to it :-(

So I would suggest to raise an issue at the Swagger project.

ricardozanini:
Hmmm last time I could infer in their parser. I'll take a look once I get the time. THat's a real shame to have such errors swallowed by the underlying library.

chberger:
That's the affected code:

```
package io.swagger.parser;

import io.swagger.v3.parser.core.extensions.SwaggerParserExtension;
import io.swagger.v3.parser.core.models.AuthorizationValue;
import io.swagger.v3.parser.core.models.ParseOptions;
import io.swagger.v3.parser.core.models.SwaggerParseResult;
import io.swagger.v3.parser.OpenAPIV3Parser;

import java.util.List;

public class OpenAPIParser {
    public SwaggerParseResult readLocation(String url, List<AuthorizationValue> auth, ParseOptions options) {
        SwaggerParseResult output = null;

        for(SwaggerParserExtension extension : OpenAPIV3Parser.getExtensions()) {
            output = extension.readLocation(url, auth, options);
            if(output != null && output.getOpenAPI() != null) {
                return output;
            }
        }

        return output;
    }
```

The variable output `` output = extension.readLocation(url, auth, options);`` points to the ``SwaggerParseResult`` which contains the error. However, in the loop the condition ``if(output != null && output.getOpenAPI() != null)`` isn't true cause ``output.getOpenAPI()`` is null and cause we have a 2nd iteration they overwriting the output instance within the next execution...

IMHO definitely something we have to report upstream ..

ricardozanini:
Yes, I agree. We are using a somewhat old version. Maybe they've fixed this in new releases. Feel free to open an issue there. :D

chberger:
I've tested it with the latest version of ``openapitools 6.4.0`` which depends on the latest ``openapi-parser``:

It seems the behavior has changed, at least I've got a ``Nullpointer`` and the test failed:
```
java.lang.NullPointerException: Cannot invoke "io.swagger.v3.oas.models.OpenAPI.getExtensions()" because "openAPI" is null
```
I could open a PR to bump the ``openapitools`` version. Only little changes are required (the models from the ``skip-validation`` test have some proper names now instead of  ``InlineObjectXX``. Anyway, haven't anaylzed it further..)

@hbelmiro wdyt?

ricardozanini:
That's a great idea. Let's upgrade it.

chberger:
> That's a great idea. Let's upgrade it.

https://github.com/quarkiverse/quarkus-openapi-generator/pull/273

PR #277, opened 2023-03-10
fix: Several GitHub actions are outdated

chberger:
Fixes the following GitHub Action warnings:

```
Node.js 12 actions are deprecated. Please update the following actions to use Node.js 16: actions/checkout@v2, actions/setup-java@v2. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/.
The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
The `save-state` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
```

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -11,3 +11,7 @@ updates:
       interval: "daily"
     ignore:
       - dependency-name: "org.apache.maven.plugins:maven-compiler-plugin"
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -34,9 +34,9 @@ jobs:
         run: git config --global core.autocrlf false
         if: startsWith(matrix.os, 'windows')
 
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - name: Set up JDK 11
-        uses: actions/setup-java@v2
+        uses: actions/setup-java@v3
         with:
           distribution: temurin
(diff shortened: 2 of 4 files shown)


.github/dependabot.yml
```diff
@@ -10,4 +10,8 @@ updates:
     schedule:
       interval: "daily"
     ignore:
-      - dependency-name: "org.apache.maven.plugins:maven-compiler-plugin"
\ No newline at end of file
+      - dependency-name: "org.apache.maven.plugins:maven-compiler-plugin"
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
```
ricardozanini, line 17: I think it's fair keeping it `daily`.

ricardozanini (commented):
Thank you!

.github/dependabot.yml
chberger: Since we use major versions at the actions only, we should get minor updates anyway. So IMHO the bot should just raise an issue once a major update is available. Thus I think a ``weekly`` interval should be fine. But but ultimately you decide ...

hbelmiro (approved):

hbelmiro:
Cherry-picked to quarkus2 branch: https://github.com/quarkiverse/quarkus-openapi-generator/pull/285

PR #291, opened 2023-03-29
fix: Alignment of the capability detection logic

chberger:
My upstream contribution has been merged https://github.com/quarkusio/quarkus/pull/31769, thus we can align the capability detection logic. 

However, I'm not sure whether we should backport this to ``quarkus2``, cause it requires at least Quarkus ``2.16.5.Final``.

@hbelmiro WDYT?


hbelmiro (approved):
Thank you for this contribution, @chberger.
I believe we shouldn't backport this to `quarkus2` branch so users can keep using earlier versions of Quarkus.

Issue #308, opened 2023-04-25
Duplicate Authorization header produced by ApiKeyAuthenticationProvider#filter (with reproducer)


laurentperez:
Hello

Using 1.3.1, I have a use case where multiple `Authorization` headers are sent resulting in a `400 Bad request` response from a strict http server.

Root cause is https://github.com/quarkiverse/quarkus-openapi-generator/blob/1.3.1/runtime/src/main/java/io/quarkiverse/openapi/generator/providers/ApiKeyAuthenticationProvider.java#L44 does `.add` and appends multiple headers withe the same `Authorization` name, to the original one in openapi definition : 

```
                    {
                        "name": "Authorization",
                        "in": "header",
                        "description": "Authorization",
                        "required": true,
                        "type": "string"
                    }
```             

Not sure how to fix this myself so I open an issue. Suggestions welcome.

Reproducer :

- enable http trace to see multiple Authorizations and set an api key

```
%test.quarkus.rest-client.fooopenapi_json.logging.scope=request-response
%test.quarkus.openapi-generator.fooopenapi_json.auth.JWT.api-key=seeee
```
- put this` fooopenapi.json` in src/main/openapi (1)

trace show the duplicated headers : 

```
2023-04-25 16:02:28,647 DEBUG [org.apa.htt.wire] (Test worker) http-outgoing-0 >> "Accept: */*[\r][\n]"
2023-04-25 16:02:28,648 DEBUG [org.apa.htt.wire] (Test worker) http-outgoing-0 >> "Authorization: seeeee[\r][\n]"
2023-04-25 16:02:28,648 DEBUG [org.apa.htt.wire] (Test worker) http-outgoing-0 >> "Authorization: seeeee[\r][\n]"
2023-04-25 16:02:28,648 DEBUG [org.apa.htt.wire] (Test worker) http-outgoing-0 >> "User-Agent: Apache-HttpClient/4.5.14 (Java/17.0.5)[\r][\n]"
2023-04-25 16:02:28,674 DEBUG [org.apa.htt.wire] (Test worker) http-outgoing-0 << "HTTP/1.1 400 Bad Request[\r][\n]"
```

questions :  

- how can I **prevent** generation of `@RegisterProvider(CompositeAuthenticationProvider.class)` and `@RegisterClientHeaders(AuthenticationPropagationHeadersFactory.class)` ?
[...]

Issue #290, opened 2023-03-27
Feature request, make it possible to override auth providers


SvenMarquardt5772:
The default generated auth providers, expect the tokens and username in properties. But I want to get them from the secrets manager of aws or gcp for example. 

To get them inside the property store is difficult, and it would be easier, if I could override the implementation of the classes `AuthenticationPropagationHeadersFactory`,`CompositeAuthenticationProvider`. 

It is difficult because, even with a custom property provider, you can't inject the `SecretsManagerClient` because property runs too early.

So, a property that I can set with the full class name would be enough for this feature.

SvenMarquardt5772:
As a workaorund, i currently use this maven plugin.

```xml
<plugin>
                <groupId>com.google.code.maven-replacer-plugin</groupId>
                <artifactId>replacer</artifactId>
                <version>1.5.3</version>
                <executions>
                    <execution>
                        <phase>prepare-package</phase>
                        <goals>
                            <goal>replace</goal>
                        </goals>
                    </execution>
                </executions>
                <configuration>
                    <basedir>${project.build.directory}/generated-sources</basedir>
                    <filesToInclude>**/*.java</filesToInclude>
                    <replacements>
                        <replacement>
                            <token>import org.openapi.quarkus.swagger_v3_v3_json.api.auth.CompositeAuthenticationProvider</token>
                            <value>import <package_name>.CompositeAuthenticationProvider</value>
                        </replacement>
                        <replacement>
                            <token>import org.openapi.quarkus.swagger_v3_v3_json.api.auth.AuthenticationPropagationHeadersFactory</token>
                            <value>import <package_name>.AuthenticationPropagationHeadersFactory</value>
                        </replacement>
                    </replacements>
                </configuration>
            </plugin>
```

jhaakma:
> As a workaorund, i currently use this maven plugin

@SvenMarquardt5772 Any idea how to get this thing to trigger when running `mvn quarkus:dev`? The damn this recompiles the module, but doesn't run the replacer even if I set the phase to compile.

SvenMarquardt5772:
@jhaakma you have to use the correct phases as well for the openapi plugin and put the configs into the executions. Here is my example.
```xml
   <plugin>
                <groupId>com.google.code.maven-replacer-plugin</groupId>
                <artifactId>replacer</artifactId>
                <version>1.5.3</version>
                <executions>
                    <execution>
                        <id>replace-auth-provider</id>
[...]

PR #311, opened 2023-04-26
#308 do not duplicate Authorization header, see https://github.com/quarkiverse/quarkus-openapi-generator/issues/308

laurentperez:
see https://github.com/quarkiverse/quarkus-openapi-generator/issues/308

this may be naive, but it works and `putSingle` is expected to guarantee *no* multiple `Authorization` headers will be sent : if multiples were sent, the receiving server would not know which one to use anyway.

- tests are passing (`mvn clean install` from a git clone of main)


hbelmiro (commented):
Thank you for this PR, @laurentperez.
`putSingle` seems to be enough.
Can you please add an integration test?

ricardozanini:
@laurentperez as @hbelmiro said, would be nice to have an integration test. I'd suggest using your own OpenAPI file in this test. See here: https://github.com/quarkiverse/quarkus-openapi-generator/tree/main/integration-tests

laurentperez:
yup. I was looking at https://github.com/quarkiverse/quarkus-openapi-generator/tree/main/integration-tests/security/src/test/java/io/quarkiverse/openapi/generator/it/security for openweather.

per RFC https://datatracker.ietf.org/doc/html/rfc7235#section-4.2 it does not explicitly says "no multiple Authorization headers are allowed" but instead "A user agent that wishes to authenticate (..) can do so by including an Authorization header field with the request.". implicitely this "an" means "only one". hence, I putSingle.

unsure yet if I should use a custom OpenApi file for the test. the test is simply a generic test to make sure only one header is sent by the `Provider`

I'll see if I can do a simple thing following the openweather IT tests.

ricardozanini:
@laurentperez adding one case test to one already in place is pretty reasonable too. Thanks!

laurentperez:
added a full IT test since it was easier than I expected using wiremock. the "fooapikey" is the expectation. @hbelmiro @ricardozanini  

I reduced the own OpenAPI file to the minimal stuff to avoid cluttering

tests are green locally (mvn clean install)

integration-tests/security/src/test/java/io/quarkiverse/openapi/generator/it/security/AuthorizationHeaderApiKeyCanFilterWithoutDuplicateAuthorizationTest.java
```diff
@@ -0,0 +1,42 @@
...
+    FooResourceApi fooResourceApi;
+
+    @Test
+    public void testNoMultipleAuthorizationHeadersAreSent() {
+        List<FooDTO> foos = fooResourceApi.getFoosUsingGET("not the fooapikey",
+                123465L);
+        assertNotNull(foos);
+        fooServer.verify(WireMock.getRequestedFor(
+                        WireMock.urlEqualTo("/api/foo/v2.0/foo?something=123465"))
+                .withHeader("Authorization", equalTo("fooapikey"))
```
ricardozanini, line 38: We might need to check if the header is not duplicated here. Isn't it the reason behind this PR?

integration-tests/security/src/main/resources/application.properties
```diff
@@ -7,6 +7,11 @@ org.acme.openapi.security.auth.basic_auth/password=test
 quarkus.openapi-generator.codegen.spec.open_weather_yaml.base-package=org.acme.openapi.weather
 quarkus.openapi-generator.open_weather_yaml.auth.app_id.api-key=12345
 
+# #308 use the configured api-key as the Authorization header
+# to avoid case of @HeaderParam("Authorization") in openapi definition creating multiple Authorization headers
```
hbelmiro, line 11: I don't think this comment is needed.
WDYT @ricardozanini?

ricardozanini: I agree.

integration-tests/security/src/test/java/io/quarkiverse/openapi/generator/it/security/AuthorizationHeaderApiKeyCanFilterWithoutDuplicateAuthorizationTest.java
ricardozanini: I see! Alternatively, you could add a new test case using RestAssured and verify the number of headers in the response.

Or maybe you might have access to https://github.com/wiremock/wiremock/blob/master/src/main/java/com/github/tomakehurst/wiremock/matching/RequestPattern.java#L496?

laurentperez: Yep. I'll 👀 into these.

Basically it's about the request headers, the response doesn't really matter.
My use case triggers a 400 Bad request given the multiple headers and I can't do the very dirty thing from https://github.com/nickstenning/nginx-multiauth

As you probably guessed, the "weird" OpenApi definition I can't change it, mine comes from a 3rd party service vendor.

It's weird, but still spec compliant afaik. One party has the right to define a required `Authorization` as a `HeaderParam`.

The "problem" is that it clashes with `quarkus.openapi-generator.fooopenapi_json.auth.JWT.api-key` : this prop is not mandatory :  yet `ApiKeyAuthenticationProvider.java` wants a key. Otherwise it'd get a null one and write the empty "Authorization: emptyness" header.

My `.putSingle` works because it is _shadowing_ either the null key, but or & also the not-null HeaderParam one : at last only one Authorization will remain : the one from the `application.properties`, fixing the use case. Unsure if it's the right call... Thoughts on this ?

- Shall the ApiKeyAuthenticationProvider write a warning or info in logs at buildtime or runtime if one forgets to declare `api-key` in .properties ?

- I was pondering about using a noop Provider which basically, provides nothing : zero headers, zero filtering, per https://github.com/quarkiverse/quarkus-openapi-generator#custom-register-providers-for-generated-api

- This noop would shadow the ApiKeyAuthenticationProvider.

ricardozanini: oh sure, it's the request we make, so yeah getting the headers from WireMock instance should solve it.

ricardozanini: > Shall the ApiKeyAuthenticationProvider write a warning or info in logs at buildtime or runtime if one forgets to declare api-key in .properties ?

Yes, we can since it's the single source of ApiKey at the moment. A WARN in runtime would make total sense.

> I was pondering about using a noop Provider which basically, provides nothing : zero headers, zero filtering

How is this would be useful? 

LOL now that I see that we already support custom providers, @hbelmiro. About the discussion at #290

laurentperez: I found a way to count the headers, added a warn log. Pushed. Tests pass.

I was thinking about a custom noop Provider simply to ignore the registration of `@RegisterProvider(CompositeAuthenticationProvider.class)` at _codegen_. Because the composite is the one pluggin the `ApiKeyAuthenticationProvider`.

This would avoid the header duplication bug by... hiding it by omission : if no ApiKeyAuthenticationProvider is registered then no Authorization headers are sent :smile_cat: 

Ignoring the registration woul be hacky. I believe my latest commits are better & ok enough.

ricardozanini (approved):
Thank you!

runtime/src/main/java/io/quarkiverse/openapi/generator/providers/ApiKeyAuthenticationProvider.java
```diff
@@ -41,13 +45,17 @@
...
+                requestContext.getHeaders().putSingle(apiKeyName, getApiKey());
                 break;
         }
     }
 
     private String getApiKey() {
-        return getAuthConfigParam(API_KEY, "");
+        final String key = getAuthConfigParam(API_KEY, "");
+        if ("".equals(key)) {
+            LOGGER.warn("configured " + API_KEY + " property (see application.properties) is empty. hint: configure it.");
```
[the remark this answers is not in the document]
hbelmiro: @laurentperez can you please just replace
```java
if ("".equals(key)) {
```

with

```java
if (key.isEmpty()) {
```

hbelmiro (approved):

hbelmiro:
@allcontributors please add @laurentperez for code

Issue #192, opened 2022-12-02
Change `OpenApiGeneratorCodeGenBase#shouldRun` to throw an exception rather than return `false`


hbelmiro:
Silently returning `false` in `io.quarkiverse.openapi.generator.deployment.codegen.OpenApiGeneratorCodeGenBase#shouldRun` may confuse the user of the extension, that may not notice that an invalid path is being used.
It will be clearer if we throw an exception instead, so the build will fail.

This should be done after https://github.com/quarkiverse/quarkus-openapi-generator/pull/179 is merged.

brunobaiano:
I have a strange behavior in integrated-tests after implementing this:
I have to create a lot of openapi folders inside src/test
Is this an expected behavior?
You can check the modification here:

https://github.com/brunobaiano/quarkus-openapi-generator/commit/2339a5a3c7135899388a569b9a9dfa05bacc9778


The other file is commented because I'm not sure how to add a provider yet. (and now I have an error too)

hbelmiro:
@brunobaiano it is actually the bug that originated this issue. See [this comment](https://github.com/quarkiverse/quarkus-openapi-generator/pull/179#discussion_r1037929730).
It happens because the tests try to find the openapi directory in src/test. It should try to find in src/main.
So, the PR to fix this issue has to change `OpenApiGeneratorCodeGenBase#shouldRun` to not return `false` without having to create the directories under src/test.

brunobaiano:
It's not returning false, it's throwing the exception. The thing is, all tests will fail now.
 I can fix adding  `quarkus.openapi-generator.codegen.input-base-dir=src/main/openapi`
on each application.properties in integration-tests. The change-directory test has this property and works.
you can see here https://github.com/brunobaiano/quarkus-openapi-generator/commit/db74b9e3f89e3743c86ac0b086467390646b5417

hbelmiro:
Sorry, @brunobaiano. I might have not expressed myself properly.

The PR to fix this issue needs to address two requirements:

1. `OpenApiGeneratorCodeGenBase#shouldRun` can't return `false` at all. It must return only `true` or throw an exception. That's what you already did.
2. The integration tests have to pass without you having to manually create the directory or set `input-base-dir=src/main/openapi` in `application.properties`. When users use this extension they shouldn't need to do that. Therefore, in integration tests, this mustn't be necessary either. Fixing it is part of the solution to this issue.

brunobaiano:
[...]

PR #314, opened 2023-04-28
 #192 Change OpenApiGeneratorCodeGenBase#shouldRun to throw an exception rather than return false

brunobaiano:
shouldRun now returns true
if sourceDir equals src/test/openapi will not throw an exception (this will not break generate-code-tests)
Improve getInputBaseDirRelativeToModule to return Optional and avoid null


hbelmiro (approved):
Looks good, @brunobaiano.
Thank you for this contribution.

hbelmiro:
@allcontributors please add @brunobaiano for code

hbelmiro:
Cherry-picked to quarkus2 branch https://github.com/quarkiverse/quarkus-openapi-generator/pull/316

Issue #352, opened 2023-05-29
feat: Allow dynamic api keys in ApiKeyAuthenticationProvider


laurentperez:
this is a follow up of the root reproducer https://github.com/quarkiverse/quarkus-openapi-generator/issues/308 and its https://github.com/quarkiverse/quarkus-openapi-generator/pull/311#issuecomment-1527313965 patch

@hbelmiro I'm opening this issue for discussion because my PR did not cover 100% of my use case

tl;dr my `putSingle` works : only one Authorization header is sent

however, the one being sent is the one from the `application.properties` : this works when the api key is known at `compile` time.

however my use case is : the api key is known at `runtime` (this is basically a JWT Bearer key)

so the `@HeaderParam String apiKey Authorization header` known at runtime has to win over the api key from the .properties. my PR did not cover this case.

I can PR this. My plan is to let the @HeaderParam override the .properties, however, this **may** break things. If people relied on .properties, well, it will be overriden.

Did you see this use case before ? This is basically a case where API endpoints do not all share the same api key, for security reasons or dispatching of requests.

hbelmiro:
We can create a new configuration to define which one has the bigger priority. The default one could be `application.properties`, so we don't break things.

What do you guys think @laurentperez @ricardozanini?

@laurentperez I'll be offline for a couple of weeks, so I might not see your next messages. @ricardozanini will be around though.

laurentperez:
> The default one could be application.properties, so we don't break things.

this is what I was thinking too. I'll draft a PR.

ricardozanini:
@laurentperez go ahead with @hbelmiro suggestion. I also think that's a good approach to not breaking things. Thanks!

PR #355, opened 2023-06-05
#352 enable api key override for Authorization header

laurentperez:
it covers the use case of #352 where API key is obtained dynamically

`application.properties` configuration has `precedence` over dynamic API key runtime discovery


laurentperez:
@ricardozanini  Hi. WDYT , is the PR good enough ? I kept things simple, simply reusing the Authorization header if it existed.

hbelmiro:
Cherry picked to quarkus2 branch: https://github.com/quarkiverse/quarkus-openapi-generator/pull/366

Issue #357, opened 2023-06-06
Adding @QueryParam to POJO fields does not work with POST/PUT/PATCH using quarkus-rest-client-reactive-jackson


magori:
In the last version [2.2.5](https://github.com/hbelmiro/quarkus-openapi-generator/blob/347f02861e35e3aa9e863bf93aa3f81106dc3f6b/deployment/src/main/resources/templates/pojo.qute#L52) the QueryParam annotation is added on the fields.
But now when we execute a PUT with the rest-client-reactive the body is empty. 

LOG with @QueryParam on field name
_ 2023-06-06 11:30:20,588 DEBUG [org.jbo.res.rea.cli.log.DefaultClientLogger] (vert.x-eventloop-thread-2) Request: PUT http://localhost:8080/api/v1/project?name=Groupe1%2Fprojettest Headers[Accept=application/json Content-Length=0 User-Agent=Resteasy Reactive Client X-Api-Key=xxx], Empty body_

LOG without @QueryParam on field name
_2023-06-06 11:35:03,264 DEBUG [org.jbo.res.rea.cli.log.DefaultClientLogger] (vert.x-eventloop-thread-5) Request: PUT http://localhost:8080/api/v1/project Headers[Accept=application/json Content-Type=application/json User-Agent=Resteasy Reactive Client X-Api-Key=xxx content-length=29], Body:
{"name":"Groupe1/projettest"}_

Is it possible to deactivate/delete this annotation? 

For information we use Kotlin for the project

ricardozanini:
Hi! Thanks for reporting it. Can you please share an OpenAPI model in which this annotation is wrongly generated?

jasmdk:
Any spec where which has a requestBody will generate pojos where each property has the QueryParam annotation, and when invoking the request, all properties are added as query parameters to the URL and leaves the body empty

ricardozanini:
I asked one because it will make my life easier to fix this bug as soon as possible. Thanks for the information anyway, I'll take care of it once I have the time.

frederikb:
> I asked one because it will make my life easier to fix this bug as soon as possible. Thanks for the information anyway, I'll take care of it once I have the time.

Hi @ricardozanini: I created a PR which enhances the BeanParam integration test with a failing case. See #361 .

[...]

PR #361, opened 2023-06-14
Add failing test for RESTEasy Reactive when using request bodies

frederikb:
Adds a test that fails due to the request body types (here: PatchRequestDto) getting generated with @QueryParam annotations on each field. This results in PATCH (or POST) requests that have each field as a query param as part of the URL ("/patch?name=Max&age=42") instead of as part of a JSON body.

This currently fails **only** if run against RESTEasy Reactive Client. The legacy RESTEasy Classic Client generates the expected body. In both cases the source code generated by the OpenAPI Generator remains the same.

To my knowledge the unwanted behavior of RESTEasy Reactive is however to spec and the @QueryParam annotation should not be added to types used as RequestBody. Of course this means that one should never share types used both for query request parameters and request bodies.

I'm opening this PR as a showcase for issue #357. Feel free to close/delete if obsolete or not needed.

### DIFF
--- a/integration-tests/beanparam/src/main/openapi/openapi.yaml
+++ b/integration-tests/beanparam/src/main/openapi/openapi.yaml
@@ -8,13 +8,13 @@ info:
     url: http://www.apache.org/licenses/LICENSE-2.0
   version: "4.0.0"
 tags:
-  - name: GetTest
-    description: GetTest Controller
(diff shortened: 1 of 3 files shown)


hbelmiro:
Closing as this is just a reproducer.

PR #386, opened 2023-07-11
Change api template to return jakarta Response instead of void

lordvlad:
Imagine POSTing to a resource that creates something and returns no
payload but a 201 with a Location header. Right now, there is no way to
get that url. After this commit, it
will be possible

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,7 +152,9 @@ When using RESTEasy Reactive:
 
 ## Returning `Response` objects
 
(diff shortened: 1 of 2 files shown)


hbelmiro (approved):
Thank you for this contribution @lordvlad.

ricardozanini (approved):

hbelmiro:
@allcontributors please add @lordvlad for code and doc

mzellho:
Hello @lordvlad , thanks for the contribution, exactly what I need and also just in time to get the latest release.

I've been playing around for quite a while and cannot get this working :-/
Can you maybe provide an example spec or a unit test on how to not define a response model?

I tried anything I could think of within the 'responses' object as well as doing some invalid things with the 'responses' object or removing 'responses' but none of them worked. Of course, I had to disable spec validation before that.

Maybe we could also add that example to the documentation for future reference.

Thank you!!

Issue #424, opened 2023-07-28
Return Response for individual Operation doesn't seem to work


mzellho:
Please see my [comment](https://github.com/quarkiverse/quarkus-openapi-generator/pull/386#issuecomment-1655446766) in #386.

hbelmiro:
@mzellho it seems to be working correctly.
I sent a [PR to test it](https://github.com/quarkiverse/quarkus-openapi-generator/pull/425).

mzellho:
Thank you, @hbelmiro, I briefly compared your spec and application.properties to mine and couldn't spot any differences right now. I'll clone your PR and see if I can find out more. Thanks, again, for your time and the new test cases!

mzellho:
Hmmm...I think I spotted the difference, @hbelmiro: In `testReturnResponseFalseVoid` you are asserting that the method should return `void` and that's either the bug or the feature. I would have thought that this test would assert `Response` instead of `void`.

I understood the feature in a way that it would also return `Response`, regardless of the state of `return-response`, so it would allow having typed Responses for all the operations featuring a response but also return `Response` in case the operation returns a 204, so Headers, Cookies, etc. can be evaluated.

What do you think?

hbelmiro:
But that's exactly what `return-response` is intended for.
If you want to return `Response` instead of `void`, set `return-response = true`.

mzellho:
Well, `return-response` is for ALL operations and my spec holds several operations, with one being void, and for all the others I want my typed Responses.

I must have misunderstood the feature, I'm afraid.

But okay, I can also take the response and deserialize manually. However, that would be a nice thing to have, maybe I'll prepare a PR that makes use of the vendor extension I mentioned in #386.

lordvlad:
My intention in #386 was to return Response instead of void if no return type was specified and return-response was default (`false`). Exactly what @mzellho was writing.

Obviously, I should've adjusted the tests as well.

What's weird to me is that where I was using it, the return types seemed to change to response, but I'll need to double check if I didn't change anything else.

So to me, the unit tests are verifying the wrong behavior, I'll try and find some time to update them.

But I guess the suggestion using the vendor extension should fix it

mzellho:
Thank you @lordvlad, I'll provide a PR around noon.

mzellho:
--> #426

hbelmiro:
[...]

PR #426, opened 2023-07-31
Response for individual operations

mzellho:
This PR fixes the return of `Response` for `void` operations (Issue #424).


integration-tests/return-response/src/main/resources/application.properties
```diff
@@ -1,4 +1,24 @@
...
+
+quarkus.openapi-generator.codegen.spec.mutiny_return_response_false_string_simple_openapi_yaml.return-response = ${quarkus.openapi-generator.codegen.spec.return_response_false_string_simple_openapi_yaml.return-response}
+quarkus.openapi-generator.codegen.spec.mutiny_return_response_false_void_simple_openapi_yaml.return-response = ${quarkus.openapi-generator.codegen.spec.return_response_false_void_simple_openapi_yaml.return-response}
+quarkus.openapi-generator.codegen.spec.mutiny_return_response_true_string_simple_openapi_yaml.return-response = ${quarkus.openapi-generator.codegen.spec.return_response_true_string_simple_openapi_yaml.return-response}
+quarkus.openapi-generator.codegen.spec.mutiny_return_response_true_void_simple_openapi_yaml.return-response = ${quarkus.openapi-generator.codegen.spec.return_response_true_void_simple_openapi_yaml.return-response}
+
+quarkus.openapi-generator.codegen.spec.mutiny_return_response_false_string_simple_openapi_yaml.mutiny = true
+quarkus.openapi-generator.codegen.spec.mutiny_return_response_false_void_simple_openapi_yaml.mutiny = true
```
hbelmiro, line 22:
```suggestion
quarkus.openapi-generator.codegen.spec.mutiny_return_response_false_string_simple_openapi_yaml.mutiny = false
quarkus.openapi-generator.codegen.spec.mutiny_return_response_false_void_simple_openapi_yaml.mutiny = false
```

integration-tests/return-response/src/test/java/io/quarkiverse/openapi/generator/it/ReturnResponseTest.java
```diff
@@ -4,8 +4,7 @@
 
 import jakarta.ws.rs.core.Response;
 
-import org.acme.openapi.api.ModelApi;
-import org.acme.openapi.api.ResponseApi;
+import org.acme.openapi.api.*;
```
hbelmiro, line 7: Please don't use star imports.
You can run `mvn verify -DskipTests` before committing to format and organize imports.

hbelmiro (changes requested):
Thank you for this contribution @mzellho.
I left a few comments and I think we also need tests for vendorExtensions.x-java-is-response-void = true.

integration-tests/return-response/src/main/resources/application.properties
hbelmiro: I mean the file is not coherent with the property value.

quarkus.openapi-generator.codegen.spec.return_response_**_true_**_string_simple_openapi_yaml.mutiny = **_false_**

mzellho: I am sorry, I again may be confused after today's work, but I don't think that `*.mutiny` has should mirror the value of `return-response`?

I basically copied the 4 existing specs and tests for the classic cases, where there are the 2 cases you added (`string` / `void`) for both options of `return-response`

- return_response_false_string_simple_openapi_yaml
- return_response_false_void_simple_openapi_yaml
- return_response_true_string_simple_openapi_yaml
- return_response_true_void_simple_openapi_yaml

and gave them reactive twins, with again the same 2 cases for `return-response`

- **mutiny**_return_response_false_string_simple_openapi_yaml
- **mutiny**_return_response_false_void_simple_openapi_yaml
- **mutiny**_return_response_true_string_simple_openapi_yaml
- **mutiny**_return_response_true_void_simple_openapi_yaml

The `*.base-package` and `*.return-response` values for each of the reactive tests mirror the values of the corresponding classic equivalent test, while the values for the `*.mutiny` flag as well as the assertions in the unit tests for `Uni` differ for these test.

hbelmiro: Oh, I just noticed these properties are to enable/disable mutiny 🤦🏽‍♂️
I was thinking they were enabling/disabling return-response. So it wasn't making any sense to me:
return_response_**_false_**_string_simple_openapi_yaml.return-response = **_true_**
Please forget all I said about this. Sorry for the confusion.

mzellho: Nevermind, thanks a lot for the review!

hbelmiro:
@mzellho I'm just waiting for a test case related to vendorExtensions.x-java-is-response-void = true.  Something like:
```java
    @Test
    void testXJavaIsResponseVoidReturnResponseFalseVoid() throws NoSuchMethodException {
        var method = XJavaIsReturnResponseVoidReturnResponseFalseVoidApi.class.getMethod("hello");
        assertThat(method.getReturnType())
                .isEqualTo(Void.TYPE);
    }
```

The rest looks good to me.

ricardozanini (approved):

mzellho:
Hi @hbelmiro,

I spent a lot of time thinking about how to test the vendor extension - long story short: I can't. As stated in my debugging notes in #386, we always end up in an `if` statement in the `AbstractJavaJAXRSServerCodegen` of `openapi-generator` if we don't specify a `response` in the spec and that's exactly where the vendor extension is added. So, even with this PR we totally rely on what `AbstractJavaJAXRSServerCodegen` has added elsewhere and likely are vulnerable to changes there.

Now, having slept over this, I really think that this PR at the very least should not be based on the vendor extension but compare for `void` (which is set as a return type in the very same if in `AbstractJavaJAXRSServerCodegen`) in our `api.qute`.

But, to be completely honest, I even think that we shouldn't add behavior that cannot be configured here. I feel that it would be a lot better and more importantly more user-friendly to revert what we did here and either:

a) provide a new config property (`boolean`) to "return `Response` if original `Response` had been void" (of course, this should be applied for mutiny as well)

or 

b)  provide a new config property (List<String>) to "name operations that should return `Response`"

Of course, the generator feature `return-response` would override these settings if set to `true`.

WDYT?

hbelmiro:
Hi @mzellho.

> Now, having slept over this, I really think that this PR at the very least should not be based on the vendor extension but compare for void (which is set as a return type in the very same if in AbstractJavaJAXRSServerCodegen) in our api.qute.

It makes sense to me.
I would go with:

> b) provide a new config property (List) to "name operations that should return Response"

Or even change the current `return-response` option to accept both `boolean` and `List`. Something like:

```properties
...return-response = true // return `Response` for all operations
...return-response = false // return type/void for all operations
...return-response = op1,op2,op3 // return `Response` only for op1, op2, and op3
```

WDYT?

mzellho:
Hey @hbelmiro, thanks for the feedback. I would also prefer (b) and am confident that there'll be some time for it before the weekend.

As `return-response` is a feature of `openapi-generator` I would prefer not to mix it up with custom behavior, but how about adding a `Map<String, String>` config property instead that allows to specify full-qualified class names as the desired response type per operation, with any undefined operation falling back to default generator behavior  (e.g. `op2` of below configuration would use the calculated response type).

```properties
quarkus.openapi-generator.codegen.spec.myspec_yaml.responses.op1=jakarta.ws.rs.core.Response # override Response
quarkus.openapi-generator.codegen.spec.myspec_yaml.responses.op3=void
```

If `mutiny` is flagged to `true`, the responses would of course be `Uni`.
If `return-response` is flagged to `true`, any specified response in the new config property will be ignored.

hbelmiro:
@mzellho `return-response` is not from openapi-generator. I implemented this feature in https://github.com/quarkiverse/quarkus-openapi-generator/pull/219.
Do you see any other drawback besides that?
A `Map<String, String>` config property would be a good solution, but I'd still prefer to stick with one single property for simplicity. Users wouldn't have to know/worry about overwriting properties.

mzellho:
@hbelmiro Just to be safe, isn't `return-response` forwarded to the jaxrs generator as an additional property?

hbelmiro:
> @hbelmiro Just to be safe, isn't `return-response` forwarded to the jaxrs generator as an additional property?

@mzellho It is. And then we [read its value in the qute template](https://github.com/quarkiverse/quarkus-openapi-generator/blob/4c4ad53f361b42d5f83dd8ae26b2c692f9153d57/deployment/src/main/resources/templates/api.qute#L96C24-L96C24), just how we do with other custom properties.

lordvlad:
> I feel that it would be a lot better and more importantly more user-friendly to revert what we did here

I don't think it is more user-friendly to have an additional option for this. I get the point that this behavior possibly diverges from standard openapi-generator behavior. But then again, since this project is using it's own qute templates, you could argue that it's already diverging from standard openapi-generator behavior. 

Correct me if I'm wrong, but I don't think this PR, as it is, would break any expectations or worsen developer experience in any way. If you don't care for the response, there's no difference in having a void method or simply discarding the return value. On the other hand, if you _do_ care for the response (and there is no type info given) its much more user-friendly to already have the `Response` object instead of having to look for ways to turn the void into a response. I don't think there is any value in keeping the `void` return type, but I'd love to be corrected.

hbelmiro:
@mzellho since @lordvlad's comment above makes a lot of sense, would you mind changing this PR to always return `Response`?

hbelmiro (approved):

hbelmiro:
https://github.com/all-contributors please add @mzellho for code
https://github.com/all-contributors please add @mzellho for test

Issue #435, opened 2023-08-10
Generated client not usable because of @PartFilename’s value - filename value is missing a supported extension


patr1kt0th:
Hi guys. I'm not sure if it really is an issue, but as I have no (good) idea how to solve it, I'll give it a shot.

We're accessing an external (PDF) conversion service and are using `quarkus-openapi-generator` to generate the REST client based on their OpenAPI YAML. We're using `resteasy-reactive` and the generated code for adding data (files) to a conversion job looks like this:

```
    @POST
    @Path("/{jobId}/data/")
    @Consumes({"multipart/form-data"})
    @Produces({"application/json", "application/xml", "application/problem+json", "application/problem+xml"})
    @GeneratedMethod ("addData")
    public DataInfo addData(
        @jakarta.ws.rs.BeanParam AddDataMultipartForm multipartForm,@GeneratedParam("jobId") @PathParam("jobId") String jobId,@GeneratedParam("url") @QueryParam("url") String url
    );

    public static class AddDataMultipartForm {
        @FormParam("options")
        @org.jboss.resteasy.reactive.PartType(MediaType.APPLICATION_JSON)
        public List<Option> options;
        @FormParam("data")
        @org.jboss.resteasy.reactive.PartFilename("dataFile")
        @org.jboss.resteasy.reactive.PartType(MediaType.APPLICATION_OCTET_STREAM)
        public File data;
    }
```

The problematic part for us is `@org.jboss.resteasy.reactive.PartFilename("dataFile")`. The conversion service rejects our requests, as it only supports a bunch of allowed extensions (txt, html, docx, ...), and unfortunately `dataFile` = a filename without an extension, is not one of them.

We've "solved" the issue by simply removing the `@PartFilename` using `maven-replacer-plugin`. The problem with this "solution" was, that it did not work all the time together with `quarkus dev` = sometimes the replacer was triggered correctly, but most of the time it wasn't. After updating both Quarkus (3.2.3.Final) and `quarkus-openapi-generator` (2.2.8) this does not seem to be the case anymore (at least after doing a few test runs), which is great, but I'd still be interested if there's a more elegant solution out there.

Would it make sense to have the `@PartFilename` optional in `multipartFormdataPojo.qute`, e.g. based on some configuration property? Or is there some other way? Thanks a lot for any suggestions.

hbelmiro:
[...]

PR #452, opened 2023-08-28
Add @PartFilename config - on/off, value, value suffix

patr1kt0th:
Fixes: https://github.com/quarkiverse/quarkus-openapi-generator/issues/435

### DIFF
--- a/README.md
+++ b/README.md
@@ -649,7 +649,7 @@ By default, the `PartFilename`'s value representing the filename is prefixed by 
 quarkus.openapi-generator.codegen.spec.my_multipart_requests_yml.use-field-name-in-part-filename=false
 ```
 
-And in case the default `PartFilename` value is not suitable (e.g. a conversion service only allows/support specific extensions), the value can be set by using the `part-filename-value` property (gloablly or corresponding to your spec) in the `application.properties`, e.g.:
+And in case the default `PartFilename` value is not suitable (e.g. a conversion service only allows/supports specific extensions), the value can be set by using the `part-filename-value` property (globally or corresponding to your spec) in the `application.properties`, e.g.:
 
 ```properties
 quarkus.openapi-generator.codegen.spec.my_first_multipart_requests_yml.part-filename-value=".pdf"
--- a/deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/CodegenConfig.java
+++ b/deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/CodegenConfig.java
@@ -51,7 +51,7 @@ public class CodegenConfig extends GlobalCodegenConfig {
         ENABLE_SECURITY_GENERATION("enable-security-generation"),
         GENERATE_PART_FILENAME("generate-part-filename"),
         PART_FILENAME_VALUE("part-filename-value"),
-        USE_FILED_NAME_IN_PART_FILENAME("use-field-name-in-part-filename");
+        USE_FIELD_NAME_IN_PART_FILENAME("use-field-name-in-part-filename");
 
         private final String name;
 
--- a/deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/codegen/OpenApiGeneratorCodeGenBase.java
+++ b/deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/codegen/OpenApiGeneratorCodeGenBase.java
@@ -192,7 +192,7 @@ public abstract class OpenApiGeneratorCodeGenBase implements CodeGenProvider {
                 .ifPresent(generator::withPartFilenameValueConfig);
 
         generator.withUseFieldNameInPartFilenameConfig(
-                getValues(config, openApiFilePath, CodegenConfig.ConfigName.USE_FILED_NAME_IN_PART_FILENAME,
+                getValues(config, openApiFilePath, CodegenConfig.ConfigName.USE_FIELD_NAME_IN_PART_FILENAME,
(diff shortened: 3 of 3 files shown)


hbelmiro:
@patr1kt0th for you to run the RESTEasy Reactive tests locally:
```shell
cd integration-tests
mvn clean install -Presteasy-reactive
mvn verify -Presteasy-reactive
```

patr1kt0th:
@hbelmiro thanks, I've already prepared a "fix" and split the test classes (classic & reactive), but I was/am fighting with a compilation error, so I'll try your suggestion this evening and let's see... will get back to you as soon as I have something new

patr1kt0th:
@hbelmiro, @ricardozanini, please check. It worked locally, so hopefully there won't be any problems.

But I have one question... I did the new config options in a similar way the others are done, but I was thinking whether it wouldn't be better to put all _PartFilename_ related config under a `part-filename` key and use/rename them for example like this:

- `quarkus.openapi-generator.codegen.(spec.xyz).partfile-name.generate=true|false`
- `quarkus.openapi-generator.codegen.(spec.xyz).partfile-name.value=<some-value>`
- `quarkus.openapi-generator.codegen.(spec.xyz).partfile-name.add-field-name-to-value=true|false` (because I don't like the `part-filename-value-suffix` that much)

What do you think?

deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/wrapper/OpenApiClientGeneratorWrapper.java
```diff
@@ -179,6 +179,21 @@ public void withTemplateDir(Path templateDir) {
...
+        this.configurator.addAdditionalProperty("generate-part-filename", generatePartFilename);
+        return this;
+    }
+
+    public OpenApiClientGeneratorWrapper withPartFilenameValueConfig(final String partFilenameValue) {
+        this.configurator.addAdditionalProperty("part-filename-value", partFilenameValue);
+        return this;
+    }
+
+    public OpenApiClientGeneratorWrapper withPartFilenameValueSuffixConfig(final String partFilenameValueSiffix) {
```
patr1kt0th, line 192: Typo, it should be `partFilenameValueSuffix` not `partFilenameValueSiffix`

integration-tests/part-filename/pom.xml
```diff
@@ -0,0 +1,93 @@
...
+      <artifactId>quarkus-openapi-generator</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.jboss.resteasy</groupId>
+      <artifactId>resteasy-multipart-provider</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>io.quarkus</groupId>
+      <artifactId>quarkus-resteasy-reactive-common</artifactId>
+    </dependency>
```
hbelmiro, line 27: We declare dependencies that are specific for RESTEasy Reactive or Classic in its equivalent profile in [integration-tests/pom.xml](https://github.com/quarkiverse/quarkus-openapi-generator/blob/5f10fa319d4b58d37ce68e6aaa953796241c11a3/integration-tests/pom.xml#L47). I'm not sure about `resteasy-multipart-provider`, but you can move at least `quarkus-resteasy-reactive-common` to there.

hbelmiro (commented):
Thank you for this contribution @patr1kt0th. Very good job!
I just left a comment regarding where you placed the dependencies for the test.

And one picky suggestion: it seems you can avoid code duplication in tests. Maybe using parameterized tests or just extracting common parts to a different method. IDK, it's up to you and totally optional. If you want to leave it as it is, that's fine for me.

patr1kt0th:
Thanks for the feedback, @hbelmiro. I fully agree with the picky suggestion... I didn't want to spend too much time with it, I assume. But I improved it a bit, please check.

I also fixed the typo and the README's format and I tried to implement the suggestion I made earlier = introducing the `part-filename` config group, but it did not work as I imagined it (it worked, but there were some warnings), so I just renamed the "suffix" config parameter.

integration-tests/part-filename/pom.xml
patr1kt0th: Thanks for the info, I noticed it, but I did not want to put them there, as they're (only) relevant to the `part-filename` tests and I didn't want to influence the rest of the tests.

I need them to be able to compile the tests as they're checking both types of annotations and I couldn't figure out, how to do it in a more elegant way. Or maybe I'm just doing something wrong?

Btw. I think the `quarkus-resteasy-reactive-common` should already be part of `quarkus-rest-client-reactive-jackson` which you already have there.

patr1kt0th:
Oh, I just noticed I messed up the README.md file. @hbelmiro, can you please tell me what kind of MD settings you use in IntelliJ? Is it documented somewhere? Thanks.

hbelmiro:
@patr1kt0th we don't have any specific MD settings. As long as you don't mess it up too much, that's fine (which doesn't seem to be the case 😆).
I'll review it ASAP.

patr1kt0th:
Thanks @hbelmiro, I reverted the messed-up formatting and hopefully fixed all the issues it caused.

ricardozanini (approved):

patr1kt0th:
Thanks for the review and the suggestions/fixes @ricardozanini 👍

patr1kt0th:
Thanks for your cooperation, guys 🤝

hbelmiro:
@all-contributors please add @patr1kt0th for code

Issue #460, opened 2023-09-04
OpenAPI parsing issues


eloycoto:
Hi, 

Checking some OpenAPI spec parser in https://github.com/kiegroup/kogito-runtimes, I was unable to use some Openapi schemas:

# Move2Kube:

https://github.com/konveyor/move2kube-api/blob/main/assets/openapi.json

Errors:
```
[ERROR] Failed to execute goal io.quarkus.platform:quarkus-maven-plugin:2.16.9.Final:dev (default-cli) on project basic-functions: Unable to execute mojo: Compilation failure:
[ERROR] /home/eloy/dev/upstream/parados/serverless-workflow-examples/basic-functions/target/generated-sources/open-api-stream/org/kie/kogito/openapi/openapi/model/Workspace.java:[46,44] cannot find symbol
[ERROR]   symbol:   class NOT_FOUND
[ERROR]   location: class org.kie.kogito.openapi.openapi.model.Workspace
[ERROR] /home/eloy/dev/upstream/parados/serverless-workflow-examples/basic-functions/target/generated-sources/open-api-stream/org/kie/kogito/openapi/openapi/model/Project.java:[46,42] cannot find symbol
[ERROR]   symbol:   class NOT_FOUND
[ERROR]   location: class org.kie.kogito.openapi.openapi.model.Project
[ERROR] /home/eloy/dev/upstream/parados/serverless-workflow-examples/basic-functions/target/generated-sources/open-api-stream/org/kie/kogito/openapi/openapi/model/Role.java:[39,39] cannot find symbol
[ERROR]   symbol:   class NOT_FOUND
[ERROR]   location: class org.kie.kogito.openapi.openapi.model.Role
[ERROR] -> [Help 1]
```

# Github API:

https://raw.githubusercontent.com/github/rest-api-description/main/descriptions/ghes-3.10/ghes-3.10.json

And the error message:

```
[ERROR] /home/eloy/dev/upstream/parados/serverless-workflow-examples/basic-functions/target/generated-sources/open-api-stream/org/kie/kogito/openapi/openapi/model/NullableIntegration.java:[64,47] illegal start of expression
[ERROR] /home/eloy/dev/upstream/parados/serverless-workflow-examples/basic-functions/target/generated-sources/open-api-stream/org/kie/kogito/openapi/openapi/model/Integration.java:[64,47] illegal start of expression
```

The generated code looks like this:

```
/**
[...]

PR #468, opened 2023-09-07
Adding github spec to IT in order to verify it compiles

fjtirado:
This test that a modified version of github spec compiles. 
The changes are related with "+1", "-1" and "*" values. 
See issue https://github.com/OpenAPITools/openapi-generator/issues/16531

### DIFF
--- a/deployment/pom.xml
+++ b/deployment/pom.xml
@@ -10,7 +10,7 @@
   <name>Quarkus - Openapi Generator - Deployment</name>
 
   <properties>
-    <version.org.openapitools>6.4.0</version.org.openapitools>
+    <version.org.openapitools>7.0.0</version.org.openapitools>
     <version.org.slf4j>2.0.9</version.org.slf4j>
     <version.com.github.jknack>4.3.1</version.com.github.jknack>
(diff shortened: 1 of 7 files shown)


integration-tests/github/src/main/openapi/ghes-3.10.json
hbelmiro, line 1: Does this file really need to be so big?

hbelmiro (commented):
@fjtirado can you please add an integration test? (you actually created a module without any test)

integration-tests/github/src/main/openapi/ghes-3.10.json
fjtirado: This is associated to this issue https://github.com/quarkiverse/quarkus-openapi-generator/issues/460 (which should be split in different issues, but thats another question) In that isse, the github file was the one that gave more compilation issues, so I took a modified version of it. 
there is not test, because we just want to test if the openapi compiles (and there is not need to add test for that)

hbelmiro (approved):

ricardozanini (approved):

Issue #499, opened 2023-09-24
“You need to add io.quarkus:quarkus-rest-client-reactive-jackson to your dependencies.” even though it’s present


patr1kt0th:
Hi guys. I was about to update our project to the latest Quarkus version (3.4.1) and got a strange error when trying to compile our `conversion-connector` (the only one using `quarkus-openapi-generator` version `2.2.10`). We use the `reactive` versions of the dependencies:

```
    <dependency>
      <groupId>io.quarkus</groupId>
      <artifactId>quarkus-resteasy-reactive-jackson</artifactId>
    </dependency>
    <dependency>
      <groupId>io.quarkus</groupId>
      <artifactId>quarkus-rest-client-reactive-jackson</artifactId>
    </dependency>
```

and after I try to compile (or run) the application using `./mvnw clean compile`, I get:

```
[ERROR] Failed to execute goal io.quarkus.platform:quarkus-maven-plugin:3.4.1:generate-code (default) on project conversion-connector: Quarkus code generation phase has failed: InvocationTargetException: You need to add io.quarkus:quarkus-rest-client-reactive-jackson to your dependencies. -> [Help 1]
```

By using the `-e` switch, the cause is shown:

```
Caused by: io.quarkus.bootstrap.prebuild.CodeGenException: You need to add io.quarkus:quarkus-rest-client-reactive-jackson to your dependencies.
    at io.quarkiverse.openapi.generator.deployment.codegen.OpenApiGeneratorCodeGenBase.trigger (OpenApiGeneratorCodeGenBase.java:103)
```

And after checking deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/codegen/OpenApiGeneratorCodeGenBase.java's history and trying the same with the current code base from `main`, the `throw` moved to line 106.

Using the `classic` versions of the dependencies:

```
    <dependency>
      <groupId>io.quarkus</groupId>
      <artifactId>quarkus-resteasy-jackson</artifactId>
    </dependency>
    <dependency>
      <groupId>io.quarkus</groupId>
      <artifactId>quarkus-rest-client-jackson</artifactId>
    </dependency>
    <dependency>
      <groupId>io.quarkus</groupId>
      <artifactId>quarkus-resteasy-multipart</artifactId>
    </dependency>
```

I'm able to compile and run the application without any issues.

@hbelmiro , @ricardozanini , do you have an idea what's going on here? I can check in more details tomorrow, but maybe you know it right away?

[...]

PR #500, opened 2023-09-27
Determine REST_CLIENT_REACTIVE_JACKSON capability properly

chberger:
This PR fixes https://github.com/quarkiverse/quarkus-openapi-generator/issues/499

Some thoughts on this PR:
- I'm wasn't able to write some automated tests. That's because the test setup would be quite complex. We would need to compile the extension with Quarkus ``4.3.1`` and would test the extension with a business app using Quarkus in a previous 3.X version. Anyhow, I was able to test the functionality manually. Up to you if this would be good enough to continue. Otherwise somebody else is most welcome to write such a test.
- This code can be reverted as soon as we drop support for Quarkus versions earlier than ``3.4.1`` 
- I did some further testing and it looks like ``quarkus-openapi-generator`` in version ``2.2.10`` does also not support the Quarkus LTS branch (e.g. ``3.2.5.Final``). At least in my test scenario, I got also a ``rest.client.reactive.jackson`` related issue. This is different from the issue https://github.com/quarkiverse/quarkus-openapi-generator/issues/499 but maybe someone else can verify that. So we can cover it in another issue.

### DIFF
--- a/deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/codegen/OpenApiGeneratorCodeGenBase.java
+++ b/deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/codegen/OpenApiGeneratorCodeGenBase.java
@@ -24,6 +24,7 @@ import java.util.stream.Collectors;
 import java.util.stream.Stream;
 import java.util.stream.StreamSupport;
 
+import org.apache.maven.artifact.versioning.DefaultArtifactVersion;
 import org.eclipse.microprofile.config.Config;
 import org.openapitools.codegen.config.GlobalSettings;
 
@@ -33,6 +34,7 @@ import io.quarkiverse.openapi.generator.deployment.wrapper.OpenApiClassicClientG
 import io.quarkiverse.openapi.generator.deployment.wrapper.OpenApiClientGeneratorWrapper;
 import io.quarkiverse.openapi.generator.deployment.wrapper.OpenApiReactiveClientGeneratorWrapper;
 import io.quarkus.bootstrap.prebuild.CodeGenException;
+import io.quarkus.builder.Version;
 import io.quarkus.deployment.Capability;
 import io.quarkus.deployment.CodeGenContext;
 import io.quarkus.deployment.CodeGenProvider;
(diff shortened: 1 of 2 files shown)


deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/codegen/OpenApiGeneratorCodeGenBase.java
```diff
@@ -132,7 +138,7 @@ public boolean trigger(CodeGenContext context) throws CodeGenException {
     }
 
     private static boolean isJacksonReactiveClientPresent(CodeGenContext context) {
-        return isExtensionCapabilityPresent(context, Capability.REST_CLIENT_REACTIVE_JACKSON);
+        return isExtensionCapabilityPresent(context, determineRestClienReactiveJacksonCapabilityId());
```
hbelmiro, line 141: Typo
```suggestion
        return isExtensionCapabilityPresent(context, determineRestClientReactiveJacksonCapabilityId());
```

deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/codegen/OpenApiGeneratorCodeGenBase.java
```diff
@@ -151,6 +157,14 @@ private static boolean isExtensionCapabilityPresent(CodeGenContext context, Stri
                 .anyMatch(capability::equals);
     }
 
+    private static String determineRestClienReactiveJacksonCapabilityId() {
```
hbelmiro, line 160: Typo
```suggestion
    private static String determineRestClientReactiveJacksonCapabilityId() {
```

hbelmiro (commented):
Thank you @chberger.
I just left a comment regarding a typo.

ricardozanini (approved):

chberger:
@hbelmiro: IMHO we must not ``backport`` this fix into the ``quarkus2`` branch. Reason for that is the way how we detect ``jackson-reactive-client`` for ``quarkus2``. There we use dependencies instead of capabilities:

https://github.com/quarkiverse/quarkus-openapi-generator/blob/quarkus2/deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/codegen/OpenApiGeneratorCodeGenBase.java#L152

hbelmiro (approved):

Issue #568, opened 2023-11-17
ApiKey/Cookie authentication issue


ssternal:
I have a specfication using a cookie for authentication. This compiles fine, but at runtime I get an error when trying to invoke an operation. The relevant part of the specification file looks like the following:
```json
{
  "openapi": "3.0.1",
  "security": [
    {
      "cookie": []
    }
  ],
  "components": {
    "securitySchemes": {
      "cookie": {
        "type": "apiKey",
        "name": "TASKLIST-SESSION",
        "in": "cookie"
      }
    }
  }
}
```

When trying to invoke an operation, the following exception is thrown:
```
2023-11-17 13:13:47,217 ERROR [io.qua.ver.htt.run.QuarkusErrorHandler] (vert.x-eventloop-thread-0) HTTP Request to /create-file-structure failed, error id: ba033723-899c-4e55-b113-7da5bb8808bc-1: jakarta.ws.rs.ProcessingException: java.lang.UnsupportedOperationException
	[...]
Caused by: java.lang.UnsupportedOperationException
	at java.base/java.util.AbstractMap.put(AbstractMap.java:209)
	at io.quarkiverse.openapi.generator.providers.ApiKeyAuthenticationProvider.filter(ApiKeyAuthenticationProvider.java:47)
	at io.quarkiverse.openapi.generator.providers.AbstractCompositeAuthenticationProvider.filter(AbstractCompositeAuthenticationProvider.java:41)
	[...]
```

When having a look at `ApiKeyAuthenticationProvider#filter` in case of cookie it tries to add the key as a cookie:
https://github.com/quarkiverse/quarkus-openapi-generator/blob/9451de81a69182cdb96bd8b0051b99921bbc558a/runtime/src/main/java/io/quarkiverse/openapi/generator/providers/ApiKeyAuthenticationProvider.java#L40-L58

This likely is expected to throw an exception, because when reading the JavaDoc for `ClientRequestContext#getCookies` it states to return a **read-only** map:
> Returns: a read-only map of cookie name (String) to Cookie

There is a test for using cookie authentication, but this test mocks the request's cookie with a simple hash map, i.e. returning a mutable map:
https://github.com/quarkiverse/quarkus-openapi-generator/blob/e622e6a450b7747751645371c5f5ac9404132776/runtime/src/test/java/io/quarkiverse/openapi/generator/providers/ApiKeyAuthenticationProviderTest.java#L89-L100

[...]

PR #569, opened 2023-11-17
Fix API key cookie authentication (#568)

ssternal:
Fixes https://github.com/quarkiverse/quarkus-openapi-generator/issues/568:
Replaces access to read-only cookies using mutable headers during API key cookie authentication. This works around a potential `UnsupportedOperationException`.


ricardozanini (commented):
Were you able to test in your environment with your example OpenAPI file? Can you use your file to write an IT for this use case?

ssternal:
I can only work on this on fridays, but yes, my locally published snapshot fixes the issue. I will look into the ITs to see if I can create/adapt one.

ricardozanini:
Can you investigate this generated file?

```
Unsatisfied dependency for type org.openapi.quarkus.array_enum_yaml.api.ArrayEnumResourceApi
```

runtime/src/main/java/io/quarkiverse/openapi/generator/providers/ApiKeyAuthenticationProvider.java
```diff
@@ -44,7 +45,8 @@ public void filter(ClientRequestContext requestContext) throws IOException {
                 requestContext.setUri(UriBuilder.fromUri(requestContext.getUri()).queryParam(apiKeyName, getApiKey()).build());
                 break;
             case cookie:
-                requestContext.getCookies().put(apiKeyName, new Cookie.Builder(apiKeyName).value(getApiKey()).build());
+                final Cookie cookie = new Cookie.Builder(apiKeyName).value(getApiKey()).build();
+                requestContext.getHeaders().add(HttpHeaders.COOKIE, cookie);
```
fjtirado, line 49: Just keep one line here
```suggestion
                requestContext.getHeaders().add(HttpHeaders.COOKIE, new Cookie.Builder(apiKeyName).value(getApiKey()).build());
```

ssternal: Sure, applied.

ricardozanini (approved):

integration-tests/cookie-authentication/pom.xml
```diff
@@ -0,0 +1,106 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
hbelmiro, line 1: You need to add this module to https://github.com/quarkiverse/quarkus-openapi-generator/blob/e622e6a450b7747751645371c5f5ac9404132776/integration-tests/pom.xml#L12

hbelmiro (changes requested):
@ssternal you need to add the `cookie-authentication` module to its parent module so the tests can run.

integration-tests/cookie-authentication/pom.xml
ssternal: Yes, you are right. Forgot to commit this change.

ssternal:
The automated build failed due to an error with the new integration test. The issue arises due to availability of two REST implementations (classic v/s reactive). This again is because of the default activated classic maven profile. During the development of the integration test I asked myself if the classic profile should be acivated per default. Since the integration tests are designed to explicitly activate it when invoking the classic integraion tests I would recommend disabling the default behavior.

hbelmiro:
@ssternal we need to run tests with RESTEasy Classic and Reactive, unless in rare situations. PTAL at https://github.com/quarkiverse/quarkus-openapi-generator/blob/main/CONTRIBUTING.md#resteasy-reactive-and-resteasy-classic.

To fix the failing tests you need to remove the dependencies from the test pom.xml. Those dependencies are handled by the parent module. You can use https://github.com/quarkiverse/quarkus-openapi-generator/blob/main/integration-tests/simple/pom.xml as an example.

integration-tests/cookie-authentication/pom.xml
```diff
@@ -0,0 +1,106 @@
...
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-resteasy-reactive</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>io.quarkus</groupId>
+            <artifactId>quarkus-resteasy-reactive-jackson</artifactId>
+        </dependency>
```
nmirasch, line 37:
```suggestion
      
```

nmirasch: @ssternal @ricardozanini @hbelmiro Following @hbelmiro instructions, removing these dependencies fixed the testing. Could you take a look here? Thanks!

integration-tests/cookie-authentication/src/test/java/io/quarkiverse/openapi/generator/it/WiremockCookieAuthentication.java
```diff
@@ -0,0 +1,41 @@
...
+        wireMockServer.start();
+
+        wireMockServer.stubFor(get(urlPathEqualTo("/v1/test"))
+                .willReturn(aResponse().withStatus(204)));
+
+        return Map.of(URL_KEY, wireMockServer.baseUrl());
+    }
+
+    @Override
+    public void inject(final TestInjector testInjector) {
```
nmirasch, line 31:
```suggestion
    public void inject(TestInjector testInjector) {
```

nmirasch: removed the final to avoid the 'Confusing overloading of methods'

hbelmiro (approved):

hbelmiro:
@all-contributors add @nmirasch for code.

Issue #536, opened 2023-10-23
Generate default @Path("/") annotation for client methods


gh-nalo:
Using the openapi-generator has caused some minor issues in our application with redirecting calls, because the trailing "/"-slash is missing at the end of requests when there is more than one path defined in the OpenApi spec.

Example 1 - For the following Json:
```
{
  "swagger": "2.0",
  "info": {
    "version": "1.1.0",
    "title": ""
  },
  "schemes": [
    "http"
  ],
  "consumes": [
    "application/json;charset=utf-8"
  ],
  "produces": [
    "application/json;charset=utf-8"
  ],
  "tags": [
    {
      "name": "tag"
    }
  ],
  "paths": {
    "/someMethod/": {
      "post": {
        "operationId": "someMethod",
        "parameters": []
      }
    }
  }
}
```

this file is generated:

```
@Path("/someMethod/")
@RegisterRestClient( configKey="someFile_json")
@GeneratedClass(value="someFile.json", tag = "Default")
@ApplicationScoped
public interface DefaultApi {

    @POST
    @GeneratedMethod ("someMethod")
    public javax.ws.rs.core.Response someMethod(
    );

}
```

here the `@POST` has the trailing slash due to "/someMethod/" on class level.

Example 2 - But in this case:
```
{
  "swagger": "2.0",
  "info": {
    "version": "1.1.0",
    "title": ""
  },
  "schemes": [
    "http"
  ],
  "consumes": [
    "application/json;charset=utf-8"
  ],
  "produces": [
    "application/json;charset=utf-8"
  ],
  "tags": [
    {
      "name": "tag"
    }
  ],
  "paths": {
    "/someMethod/": {
      "post": {
        "operationId": "someMethod",
        "parameters": []
      }
    },
    "/someMethod/{id}": {
      "post": {
        "operationId": "someMethodId",
        "parameters": []
      }
    }
  }
}
```

which creates

```
@Path("/someMethod")
@RegisterRestClient( configKey="someFile_json")
@GeneratedClass(value="someFile.json", tag = "Default")
@ApplicationScoped
public interface DefaultApi {

    @POST
    @GeneratedMethod ("someMethod")
    public javax.ws.rs.core.Response someMethod(
    );

    @POST
    @Path("/{id}")
    @GeneratedMethod ("someMethodId")
    public javax.ws.rs.core.Response someMethodId(
    );
}
```

[...]

PR #585, opened 2023-12-11
[issue: #536] - Added annotation @Path("/") when operation is empty

jeffersoncleyson:

Include @Path("/")  when the operation is empty

- According to the issue in some applications the client method without annotation @Path("/") causes 404 NOT_FOUND.

Fixes: https://github.com/quarkiverse/quarkus-openapi-generator/issues/536

### DIFF
--- a/deployment/src/main/resources/templates/libraries/microprofile/api.qute
+++ b/deployment/src/main/resources/templates/libraries/microprofile/api.qute
@@ -58,6 +58,8 @@ public interface {classname} {
     @{op.httpMethod}
     {#if op.subresourceOperation}
     @Path("{op.path}")
+    {#else}
+    @Path("/")
     {/if}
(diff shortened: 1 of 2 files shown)


ricardozanini (commented):
Many thanks for this PR!

Can you please add a test case to verify the reported openAPI? https://github.com/quarkiverse/quarkus-openapi-generator/issues/536

hbelmiro (commented):
Thank you for this PR @jeffersoncleyson.
Besides @ricardozanini's comment above, can you please check the failing tests?

mcruzdev:
Hi @ricardozanini and @hbelmiro, I believe that this change could potentially cause breaking changes. Would making the default "/" optional be a viable solution? Alternatively, if we prefer not to introduce another configuration, we could consider accepting this approach.

PR #588, opened 2023-12-12
Do not add blank Bearer token

martinoneutrino:
Request for `BearerAuthenticationProvider` not to add a blank token to the `Authorization` header, as doing so will never satisfy any real authorization requirement, and may cause a request failure when multiple `Authorization` headers are sent due to multiple (optional) `security` schemes are defined on an endpoint.

This change could be extended to the other `AuthProvider`s, but I have not been using those.

Fixes: #589


hbelmiro (approved):
Do you mind adding an IT

ricardozanini (approved):

hbelmiro:
@all-contributors add @martinoneutrino to code.

Issue #633, opened 2024-01-22
Do not include null value for non required fields


fjtirado:
When a not required field is not set, a null value for that field is included in the request. In order to make the request smaller, we can skip that by adding an annotation in the generated code

michalsomora:
Event though i partially agree with @fjtirado imho i think this feature should be configurable and we shouldnt force the users to skip serialization of null fields since there is a standard way how to do it in quarkus:
https://quarkus.io/guides/all-config#quarkus-jackson_quarkus-jackson-serialization-inclusion
@hbelmiro could you please share your thoughts on this

michalsomora:
@ricardozanini maybe you have a time to think about my comment hm?

mcruzdev:
Leaving my opinion here, I think it should be configurable

fjtirado:
This affects the payload that is sent by the generated code to the REST server. 
Since we are using POJOs to generate the request and most optional fields are never set by users, the request was essentially a lot of unneeded properties set to null, which kills the purpose of the optional flag. 
So not sending them seems the right thing to do
Is there any use case where the payload should include an optional field set to null? I can only think of some servers where setting a value to null has semantic value (for example deleting an array when the request include a null for that array property). For those ones, I think that, rather than configuration, we should look at if the field is described as nullable in the OPen file. See nullable section [here](https://swagger.io/docs/specification/describing-parameters/). If it is nullable, then null should not be skipped for that property.

ricardozanini:
@fjtirado I agree that optional fields should not be forced to null. Otherwise, it smells anti-pattern.

On the other hand, I remember someone in this community raising a concern that a server won't accept a request if an optional field wasn't sent as part of the request (of course, again, a case where the server was badly implemented).

That being said, what should be **optional**, as in a configuration option, is the ability to force set to null non-required fields to address these edge cases. So this PR can be as is, but we should add a property to revert to the original behavior.

fjtirado:
For those cases, they added nullable to the open api spec. I think we need another PR to handle nullable. And if the openapi spec file does not match the server behaviour (does not have the nullable and the server behave as it was there), then, as it is probably done in other cases, the file should be edited locally to adjust it to the server behaviour (which is kind of equivalent to add a local property to enforce sending optional fields set to null, with the different that the null will be sent only for the nullable fields)

ricardozanini:
[...]

Issue #103, opened 2022-07-07
Remove imports from .qute templates to avoid compilation error due to ambiguity


hbelmiro:
Currently the .qute templates have fixed imports, like [these](https://github.com/quarkiverse/quarkus-openapi-generator/blob/41e09eb8f460f82f885f3de2b2c0f61096c7fc81/deployment/src/main/resources/templates/api.qute#L7).
This can lead to compilation errors due to ambiguity when a type in the OpenAPI document has the same simple name as an imported class. For instance:

```java
import org.kie.kogito.openapi.openapi.model.Response;
import javax.ws.rs.core.Response;
...
    @POST
    @Produces({"application/json"})
    @GeneratedMethod ("rootPost")
    public Response rootPost(
        CloudEvent cloudEvent
    );
```

```
Compilation error:
    public Response rootPost(
           ^ Response is ambiguous
```

To avoid this kind of issues, we could remove all the fixed imports from the .qute templates and qualify the class names on their usages. So, the api.qute template would be similar to:

```java
package {package};

{#for imp in imports}
import {imp.import};
{/for}

{#if appName}
/**
  * {appName}
  * {#if appDescription}<p/>{appDescription}{/if}
  */
{/if}
@javax.ws.rs.Path("{#if useAnnotatedBasePath}{contextPath}{/if}{commonPath}")
@org.eclipse.microprofile.rest.client.inject.RegisterRestClient({#if defaultServerUrl}baseUri="{defaultServerUrl}",{/if} configKey="{quarkus-generator.openApiSpecId}")
@io.quarkiverse.openapi.generator.annotations.GeneratedClass(value="{openapi:parseUri(inputSpec)}", tag = "{baseName}")
{#if hasAuthMethods}
@org.eclipse.microprofile.rest.client.annotation.RegisterProvider({package}.auth.CompositeAuthenticationProvider.class)
@org.eclipse.microprofile.rest.client.annotation.RegisterClientHeaders({package}.auth.AuthenticationPropagationHeadersFactory.class)
{/if}
@javax.enterprise.context.ApplicationScoped
public interface {classname} {

    {#for op in operations.operation}
    {#if op.summary}
    /**
     * {op.summary}
     *
     {#if op.notes}
     * {op.notes}
     *
     {/if}
     */
    {/if}
    @{op.httpMethod}
    {#if op.subresourceOperation}
    @javax.ws.rs.Path("{op.path}")
    {/if}
[...]

PR #634, opened 2024-01-22
Do not include null value in request

fjtirado:
Fixes https://github.com/quarkiverse/quarkus-openapi-generator/issues/633

### DIFF
--- a/deployment/src/main/resources/templates/libraries/microprofile/pojo.qute
+++ b/deployment/src/main/resources/templates/libraries/microprofile/pojo.qute
@@ -73,6 +73,9 @@ public class {m.classname} {#if m.parent}extends {m.parent}{/if}{#if m.serializa
     **/
     {#if !withXml}
     @JsonProperty("{v.baseName}")
+       {#if !v.required} 
+          @com.fasterxml.jackson.annotation.JsonInclude(com.fasterxml.jackson.annotation.JsonInclude.Include.NON_NULL)
+       {/if}
     {/if}
     {#for ext in v.vendorExtensions.x-extra-annotation.orEmpty}
     {ext}


dmartinol (commented):
lgtm, thanks!

deployment/src/main/resources/templates/libraries/microprofile/pojo.qute
```diff
@@ -9,6 +9,8 @@ import jakarta.xml.bind.annotation.XmlEnumValue;
 {#else}
 import java.lang.reflect.Type;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonInclude.Include;
```
hbelmiro, line 13: Rather than adding new imports, we've been referencing the FQCN to avoid conflicts with existing classes that may have the same name.

deployment/src/main/resources/templates/libraries/microprofile/pojo.qute
```diff
@@ -73,6 +75,9 @@ public class {m.classname} {#if m.parent}extends {m.parent}{/if}{#if m.serializa
     **/
     {#if !withXml}
     @JsonProperty("{v.baseName}")
+       {#if !v.required} 
+          @JsonInclude(Include.NON_NULL)
```
hbelmiro, line 79: Please use FQCN here (Related to the comment above).

ricardozanini (approved):

deployment/src/main/resources/templates/libraries/microprofile/pojo.qute
fjtirado: @hbelmiro All the xml classes are imported and to be honest I do not think thats a good policy, since it makes the generated class pretty difficult to read. If there is a conflict (very unlikely anyway), we will see it during ci cycle.

hbelmiro: I know there are several imports, but we've been avoiding adding new ones. There's an open [issue to remove the existing imports](https://github.com/quarkiverse/quarkus-openapi-generator/issues/103).
The conflict that may happen is with a class from the OpenAPI document. Imagine it has a custom type called `Include` defined in the OpenAPI document. The CI wouldn't get these cases.
It's very unlikely, but [it happened once](https://github.com/quarkiverse/quarkus-openapi-generator/pull/102).

mcruzdev (approved):

hbelmiro (approved):

fjtirado:
@hbelmiro can we release a new micro version?

PR #746, opened 2024-07-11
[main] Add possibility of generate code with bean validation annotationin place.

FelixScherzinger:
Add bean validation option for api and models. Adapt templates for correct bean validation usage. Add integration test for bean validation.

<details>
The current version of the generator does not support configuration of bean validation. The possibility is given in the basic java code generation. This PR introduces the option to enable this feature as well as introduce some changes to the existing templates for supporting the correct generation of annotations.
</details>

PR #771, opened 2024-08-15
Upgrade org.openapitools to 7.8.0

mcruzdev:

- [ ] You have read the [contributors guide](CONTRIBUTING.md)
- [ ] Your code is properly formatted according to [our code style](CONTRIBUTING.md#code-style)
- [ ] Pull Request title contains the target branch if not targeting main: `[0.9.x] Subject`
- [ ] Pull Request contains description of the issue
- [ ] Pull Request does not include fixes for issues other than the main ticket

<details>
<summary>
How to backport a pull request to a different branch?
</summary>

In order to automatically create a **backporting pull request** please add one or more labels having the following format `backport-<branch-name>`, where `<branch-name>` is the name of the branch where the pull request must be backported to (e.g., `backport-quarkus2` to backport the original PR to the `quarkus2` branch).

> **NOTE**: **backporting** is an action aiming to move a change (usually a commit) from a branch (usually the main one) to another one, which is generally referring to a still maintained release branch. Keeping it simple: it is about to move a specific change or a set of them from one branch to another.

Once the original pull request is successfully merged, the automated action will create one backporting pull request per each label (with the previous format) that has been added.

If something goes wrong, the author will be notified and at this point a manual backporting is needed.

> **NOTE**: this automated backporting is triggered whenever a pull request on `main` branch is labeled or closed, but both conditions must be satisfied to get the new PR created.
</details>

### DIFF
--- a/client/deployment/pom.xml
+++ b/client/deployment/pom.xml
@@ -11,7 +11,7 @@
   <name>Quarkus - Openapi Generator - Client - Deployment</name>
 
   <properties>
-    <version.org.openapitools>7.7.0</version.org.openapitools>
+    <version.org.openapitools>7.8.0</version.org.openapitools>
     <version.org.slf4j>2.0.16</version.org.slf4j>
     <version.com.github.jknack>4.3.1</version.com.github.jknack>
     <version.io.swagger.parser>2.1.22</version.io.swagger.parser>
--- a/client/deployment/pom.xml
+++ b/client/deployment/pom.xml
@@ -11,7 +11,7 @@
   <name>Quarkus - Openapi Generator - Client - Deployment</name>
 
   <properties>
-    <version.org.openapitools>7.6.0</version.org.openapitools>
+    <version.org.openapitools>7.7.0</version.org.openapitools>
     <version.org.slf4j>2.0.16</version.org.slf4j>
     <version.com.github.jknack>4.3.1</version.com.github.jknack>
     <version.io.swagger.parser>2.1.22</version.io.swagger.parser>
--- a/client/deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/wrapper/QuarkusJavaClientCodegen.java
+++ b/client/deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/wrapper/QuarkusJavaClientCodegen.java
@@ -53,7 +53,7 @@ public class QuarkusJavaClientCodegen extends JavaClientCodegen {
         this.testFolder = "";
         this.embeddedTemplateDir = "templates";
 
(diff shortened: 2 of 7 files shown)


mcruzdev:
I need to investigate a little more to remove the import statement for `@Valid` (jakarta.validation) annotation.

mcruzdev:
Hi @ricardozanini @hbelmiro, was necessary to add `jakarta.validation:jakarta.validation-api` to `resteasy-reactive` profile. 
It would be interesting for us to do a validation using capabilities just like we do with jackson but for bean validation `hibernate-validator`, WDYT?

hbelmiro (approved):
@mcruzdev many thanks for working on this.

> It would be interesting for us to do a validation using capabilities just like we do with jackson but for bean validation hibernate-validator, WDYT?

Sounds good to me.

> We added as default use-bean-validation as true, I think we need to change it to false and if the user wants to use bean validation, must add a Capability.

Why do you think we need to change it to `false`?

mcruzdev:
> @mcruzdev many thanks for working on this.
> 
> 
> 
> > It would be interesting for us to do a validation using capabilities just like we do with jackson but for bean validation hibernate-validator, WDYT?
> 
> 
> 
> Sounds good to me.
> 
> 
> 
> > We added as default use-bean-validation as true, I think we need to change it to false and if the user wants to use bean validation, must add a Capability.
> 
> 
> 
> Why do you think we need to change it to `false`?

Hi @hbelmiro, it depends haha

Combining the capability rule with use-bean-validation true as default we will break some users.

Turning it to false, the user needs to specify and add the hibernate-validator... the my answer is yes (change use bean validation to false by default)

hbelmiro:
> > @mcruzdev many thanks for working on this.
> > > It would be interesting for us to do a validation using capabilities just like we do with jackson but for bean validation hibernate-validator, WDYT?
> > 
> > 
> > Sounds good to me.
> > > We added as default use-bean-validation as true, I think we need to change it to false and if the user wants to use bean validation, must add a Capability.
> > 
> > 
> > Why do you think we need to change it to `false`?
> 
> Hi @hbelmiro, it depends haha
> 
> Combining the capability rule with use-bean-validation true as default we will break some users.
> 
> Turning it to false, the user needs to specify and add the hibernate-validator... the my answer is yes (change use bean validation to false by default)

I see. But I don't think we can default it to false. See this comment: https://github.com/quarkiverse/quarkus-openapi-generator/pull/746#pullrequestreview-2221411151
I think it's ok a small compatibility break in this case. The solution will be quite simple (add the dependency).

mcruzdev:
@hbelmiro makes sense!

ricardozanini (approved):

PR #810, opened 2024-09-26
Adding NotNull annotation for required fields

michalkolenda:
Following changes concern generating POJOs using client with `use-bean-validation` property set to `true`.

There is a functionality of annotating required fields with Jakarta @NotNull. According to my tests, it does not work in case when the field has no other constraints (like minLength, maxLength, etc.). I expect that required fields are annotated with NotNull no matter it has any other constraints or not.

I would suggest a little modification in pojo.qute in order to annotate every required field.

### DIFF
--- a/client/integration-tests/bean-validation/src/test/java/io/quarkiverse/openapi/generator/it/BeanValidationTest.java
+++ b/client/integration-tests/bean-validation/src/test/java/io/quarkiverse/openapi/generator/it/BeanValidationTest.java
@@ -6,6 +6,7 @@ import java.lang.annotation.Annotation;
 import java.lang.reflect.Field;
 import java.lang.reflect.Method;
 import java.util.Arrays;
+import java.util.stream.Stream;
 
 import jakarta.validation.Valid;
 import jakarta.validation.constraints.DecimalMax;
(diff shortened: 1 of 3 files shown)


client/integration-tests/bean-validation/src/test/java/io/quarkiverse/openapi/generator/it/BeanValidationTest.java
```diff
@@ -43,9 +44,10 @@ void testValidationAnnotationsAreInPlaceApi() {
     void testValidationAnnotationsAreInPlaceModel() throws Exception {
         Field id = ValidatedObject.class.getDeclaredField("id");
         Field name = ValidatedObject.class.getDeclaredField("name");
+        Field secondName = ValidatedObject.class.getDeclaredField("secondName");
         Field size = ValidatedObject.class.getDeclaredField("size");
 
-        assertThat(Arrays.stream(ValidatedObject.class.getFields())
+        assertThat(Stream.of(id, name, secondName, size)
```
hbelmiro, line 50: Just to be sure, have you changed this only for performance reasons?

hbelmiro (approved):

client/integration-tests/bean-validation/src/test/java/io/quarkiverse/openapi/generator/it/BeanValidationTest.java
michalkolenda: Actually not.

`getFields` method returns only `public` fields and the generated class consists only of private ones. It makes `ValidatedObject.class.getFields()` always returning an empty array.
By this reason assertion `assertThat(Arrays.stream(ValidatedObject.class.getFields())` has no chance to fail.
`getDeclaredFields()` method would be more appropriate, but I decided to utilize the objects returned by `getField` a few lines above.

I've just added the same change in the second test method.

hbelmiro (approved):

fjtirado (approved):

ricardozanini:
@allcontributors add @michalkolenda for code

PR #911, opened 2024-12-12
Draft - Feature/equals hashcode in client generation

denvitaharen:
This is an draft to discuss, its lacks test, the formatting is a bit off and can be made better, but its up for discussion.

I missed a generated equals and dashcode in my models, so I thought that I tried adding it. I think I got all cases checked, but before I add more time in this it should be discussed.

1.  Do we want this? Other Openapi client generators generates this (so it breaks stuff if we change to this generator)
2. If we want it, should it always generate methods and not being able do disable? Or should you have to enable it? Or should it be enabled by default with possibility to disabled?

This is an example of how the generated method looks:

```
    @Override
    public boolean equals(Object o) {
        if (this == o) return true;
        if (o == null || getClass() != o.getClass()) return false;

        Animal model = (Animal) o;

        return Objects.equals(born, model.born) &&
        Objects.equals(deceased, model.deceased);
    }

     @Override
     public int hashCode() {
         return Objects.hash(born,
         deceased);
     }
```


denvitaharen:
Sorry, I accidentally pushed an branch with commits from my last PR, its only this commits that's https://github.com/quarkiverse/quarkus-openapi-generator/pull/911/commits/447bb461947be44bfac596920252ef221eefeaa0

I will try and fix the branch tomorrow.

ricardozanini:
I don't see a problem adding this by default, but we should add the possibility to disable it since our hash/equals implementation might introduce problems in some projects. People are creative.

ricardozanini:
@denvitaharen why did you close this PR?

denvitaharen:
> @denvitaharen why did you close this PR?

@ricardozanini The truth is that with my (in my opinion) basic Git skills, It took to long to remove the commits that from an old branch, so I simply just made an new branch from main and added the few rows of code to the new branch. It was the most time efficient solution for me :)

PR #920, opened 2024-12-18
Added so equals and hashcode is generated in models.

denvitaharen:

- [x] You have read the [contributors guide](CONTRIBUTING.md)
- [x] Your code is properly formatted according to [our code style](CONTRIBUTING.md#code-style)
- [x] Pull Request title contains the target branch if not targeting main: `[0.9.x] Subject`
- [x] Pull Request contains description of the issue
- [x] Pull Request does not include fixes for issues other than the main ticket

I have added so `equals` and `hashCode` is generated to the models. It's possible to disabled it if someone doesn't want it.

@ricardozanini I created an new branch and pull request, it was easier and faster for me to do that.

The result looks like this:

```
    @Override
    public boolean equals(Object o) {
        if (this == o) return true;
        if (o == null || getClass() != o.getClass()) return false;

        Animal model = (Animal) o;

        return Objects.equals(born, model.born) &&
        Objects.equals(deceased, model.deceased);
    }

    @Override
    public int hashCode() {
        return Objects.hash(born,
        deceased);
    }
```

### DIFF
--- a/client/deployment/src/main/resources/templates/libraries/microprofile/pojo.qute
+++ b/client/deployment/src/main/resources/templates/libraries/microprofile/pojo.qute
@@ -1,5 +1,6 @@
 {@org.openapitools.codegen.CodegenModel m}
 import com.fasterxml.jackson.annotation.JsonProperty;
+import java.util.Objects;
 
 {#if m.description}
 /**
@@ -134,6 +135,48 @@ public class {m.classname} {#if m.parent}extends {m.parent}{/if}{#if serializabl
         sb.append("}");
         return sb.toString();
     }
+    {#if equals-hashcode}
+    {#if m.vars.size > 0}
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
+
+        {m.classname} model = ({m.classname}) o;
+
+        {#if m.vars.size == 1}
+        return Objects.equals({m.vars.0.name}, model.{m.vars.0.name});
+        {#else}
+        {#for v in m.vars}
+        {#if v_isFirst}
+        return Objects.equals({v.name}, model.{v.name}) &&
+        {#else if v_isLast}
+        Objects.equals({v.name}, model.{v.name});
+        {#else}
+        Objects.equals({v.name}, model.{v.name}) &&
+        {/if}
+        {/for}
+        {/if}
+    }
+
+    @Override
+    public int hashCode() {
+    {#if m.vars.size == 1}
+    return Objects.hash({m.vars.0.name});
+    {#else}
+        {#for v in m.vars}
+        {#if v_isFirst}
+        return Objects.hash({v.name},
+        {#else if v_isLast}
+        {v.name});
+        {#else}
+        {v.name},
+        {/if}
+        {/for}
+    {/if}
+    }
+    {/if}
+    {/if}
 
     /**
      * Convert the given object to string with each line indented by 4 spaces
--- a/client/deployment/src/main/resources/templates/libraries/microprofile/pojo.qute
+++ b/client/deployment/src/main/resources/templates/libraries/microprofile/pojo.qute
@@ -135,6 +135,7 @@ public class {m.classname} {#if m.parent}extends {m.parent}{/if}{#if serializabl
         sb.append("}");
         return sb.toString();
     }
+
     {#if equals-hashcode}
     {#if m.vars.size > 0}
     @Override
--- a/client/deployment/src/main/resources/templates/libraries/microprofile/pojo.qute
+++ b/client/deployment/src/main/resources/templates/libraries/microprofile/pojo.qute
@@ -1,6 +1,5 @@
 {@org.openapitools.codegen.CodegenModel m}
 import com.fasterxml.jackson.annotation.JsonProperty;
-import java.util.Objects;
 
 {#if m.description}
 /**
@@ -156,15 +155,15 @@ public class {m.classname} {#if m.parent}extends {m.parent}{/if}{#if serializabl
         {m.classname} model = ({m.classname}) obj;
 
         {#if m.vars.size == 1}
-        return Objects.equals({m.vars.0.name}, model.{m.vars.0.name});
+        return java.util.Objects.equals({m.vars.0.name}, model.{m.vars.0.name});
         {#else}
         {#for v in m.vars}
         {#if v_isFirst}
-        return Objects.equals({v.name}, model.{v.name}) &&
+        return java.util.Objects.equals({v.name}, model.{v.name}) &&
         {#else if v_isLast}
-        Objects.equals({v.name}, model.{v.name});
+        java.util.Objects.equals({v.name}, model.{v.name});
         {#else}
-        Objects.equals({v.name}, model.{v.name}) &&
+        java.util.Objects.equals({v.name}, model.{v.name}) &&
         {/if}
         {/for}
         {/if}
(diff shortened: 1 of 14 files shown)


docs/modules/ROOT/pages/includes/equals-hashcode.adoc
```diff
@@ -0,0 +1,9 @@
+`Hashcode` and `equals` are by default generated. It adds all the variables in the model
+to the model to the methods.
```
ricardozanini, line 2:
```suggestion
`Hashcode` and `equals` are by default generated. It adds all the variables in the model to the model to the methods.
```
I am not sure if I understood this sentence. 🤔

denvitaharen: In my head it was 100% clear when I pushed it :D I will try and refrase it! :)

denvitaharen:
```
    /**
     * Compares this object to the specified object. The result is
     * {@code true} if and only if the argument is not
     * {@code null} and is a {@code Animal} object that
     * contains the same values as this object.
     *
     * @param   obj   the object to compare with.
     * @return  {@code true} if the objects are the same;
     *          {@code false} otherwise.
     **/
    @Override
    public boolean equals(Object obj) {
        if (this == obj) return true;
        if (obj == null || getClass() != obj.getClass()) return false;

        Animal model = (Animal) obj;

        return Objects.equals(born, model.born) &&
        Objects.equals(deceased, model.deceased);
    }

    /**
     * Returns a hash code for a {@code Animal}.
     *
     * @return a hash code value for a {@code Animal}.
     **/
    @Override
    public int hashCode() {
        return Objects.hash(born,
        deceased);
    }
```

I realized that I hadn't written any javadoc for the new methods, so I added it now. It took some inspiration from the String and Long javadoc and rewrote so it fits this in.

docs/modules/ROOT/pages/includes/equals-hashcode.adoc
denvitaharen: I rewrote the documentation, I hope this ones better.

mbard:
Excellent, I really need equal/hashcode

ricardozanini (approved):
LGTM!

client/integration-tests/equals-hashcode/src/test/java/io/quarkiverse/openapi/generator/it/VerifyGenerationTest.java
```diff
@@ -0,0 +1,47 @@
...
+
+        Class<?> declaringClass = equalsMethod.getDeclaringClass();
+
+        assertEquals(declaringClass, expectedDeclaringClass);
+    }
+
+    private static Stream<Arguments> provideStringsForIsBlank() {
+        return Stream.of(
+                Arguments.of(Animal.class, "equals", Object.class),
+                Arguments.of(Animal.class, "hashCode", Object.class),
```
hbelmiro, line 32: Do we really need to check `Object`?

client/deployment/src/main/resources/templates/libraries/microprofile/pojo.qute
```diff
@@ -1,5 +1,6 @@
 {@org.openapitools.codegen.CodegenModel m}
 import com.fasterxml.jackson.annotation.JsonProperty;
+import java.util.Objects;
```
hbelmiro, line 3: Instead of importing, can you please use the FQCN to avoid compilation errors due to possible ambiguities?

docs/modules/ROOT/pages/includes/equals-hashcode.adoc
```diff
@@ -0,0 +1,9 @@
+`hashcode` and `equals` are by default generated to all models. `hashcode` and `equals` is based on all the variables
+in the model and can't be customized.
```
mcruzdev, line 2:
```suggestion
By default, `hashcode` and `equals` methods are automatically generated for all models. These methods are based on all the variables in the model and can not be customized.
```

client/deployment/src/main/resources/templates/libraries/microprofile/pojo.qute
denvitaharen: > Instead of importing, can you please use the FQCN to avoid compilation errors due to possible ambiguities?

Ofc, I will change so it uses FQCN instead. I tried to match the existing patterns and StringBuilder doesn't have use FQCN.

client/integration-tests/equals-hashcode/src/test/java/io/quarkiverse/openapi/generator/it/VerifyGenerationTest.java
denvitaharen: I thought a long time of how to verify that we don't generate `equals`and ´hashCode` and the best way I thought out was to get what class that declares a method. This way I can check if the generated model has or does not have the methods based on what set in the `application.properties`.

In the case of the models that doesn't have a generated `equals` and `hashCode`, the methods are declared in ´Object´.

What I can do is I can split the tests and don't check against `Object` and instead just check that the methods its not declared in the generated model. Or I can leave it like this?

Or does somebody have an better way of verifying the generation?

docs/modules/ROOT/pages/includes/equals-hashcode.adoc
mcruzdev: Nice PR!

client/integration-tests/equals-hashcode/src/test/java/io/quarkiverse/openapi/generator/it/VerifyGenerationTest.java
hbelmiro: My point is that `Object` will always have `equals` and `hashCode` declared. The verification will always pass.

client/deployment/src/main/resources/templates/libraries/microprofile/pojo.qute
hbelmiro: Just to complement, we still have some imports, like `com.fasterxml.jackson.annotation.JsonProperty` in the above line. But the idea is to remove them and not add more.

client/integration-tests/equals-hashcode/src/test/java/io/quarkiverse/openapi/generator/it/VerifyGenerationTest.java
denvitaharen: Thats a valid point @hbelmiro, I now see the problem with my test. I have rewritten the test so it only tests the models instead.

client/integration-tests/equals-hashcode/src/test/java/io/quarkiverse/openapi/generator/it/VerifyGenerationTest.java
```diff
@@ -1,42 +1,37 @@
...
-                        org.acme.equals.hashcode.model.Animal.class),
-                Arguments.of(org.acme.equals.hashcode.model.Animal.class, "hashCode",
-                        org.acme.equals.hashcode.model.Animal.class));
+    @Test
+    void verifyThatGeneratedModelDoesHaveMethods() {
+        var equalsMethod = getMethod(org.acme.non.equals.hashcode.model.Animal.class, "equals");
+        var hashCodeMethod = getMethod(org.acme.non.equals.hashcode.model.Animal.class, "hashCode");
+
+        assertNotEquals(equalsMethod.getDeclaringClass(), org.acme.equals.hashcode.model.Animal.class);
+        assertNotEquals(hashCodeMethod.getDeclaringClass(), org.acme.equals.hashcode.model.Animal.class);
```
hbelmiro, line 31: I think the method names are inverted.

hbelmiro (approved):

mcruzdev (approved):

fjtirado (approved):

PR #992, opened 2025-02-17
[Fix #991] Changing URLPatternMatcher implementation

### LINKED ISSUE #991: URLPatternMatcher not matching non trivial file paths
For the following operation
```
paths:
  /repos/{ref}:
    get:
      operationId: getPath
      parameters: 
        - $ref: '#/components/parameters/git-ref-only'
      responses:
        "200":
          description: OK
          content: 
            application/json:
              schema: 
                type: object
                properties:
                  product:
                    format: float
                    type: number 
```

and parameter
```
components:
  parameters:
    git-ref-only:
      name: ref
      description: The Git reference. For more information, see "[Git References](https://git-scm.com/book/en/v2/Git-Internals-Git-References)"
        in the Git documentation.
      in: path
      required: true
      example: heads/feature-a
      schema:
        type: string
      x-multi-segment: true
```

Ir is expected that URLPatternMatchers recognizes `repos/prefix/file.txt` as valid path, but thats not the case, which prevents security scheme to be applied to this operation.

fjtirado:
Fix https://github.com/quarkiverse/quarkus-openapi-generator/issues/991

### DIFF
--- a/client/runtime/src/test/java/io/quarkiverse/openapi/generator/providers/UrlPatternMatcherTest.java
+++ b/client/runtime/src/test/java/io/quarkiverse/openapi/generator/providers/UrlPatternMatcherTest.java
@@ -16,6 +16,13 @@ class UrlPatternMatcherTest {
         Assertions.assertTrue(pattern.matches(requestPath));
     }
 
+    @ParameterizedTest
+    @MethodSource("providePathsThatNotMatch")
+    void verifyPathsNotMatch(final String pathPattern, final String requestPath) {
+        UrlPatternMatcher pattern = new UrlPatternMatcher(pathPattern);
+        Assertions.assertFalse(pattern.matches(requestPath));
+    }
+
     private static Stream<Arguments> providePathsThatMatch() {
         return Stream.of(
                 Arguments.of("/pets/{id}", "/pets/1"),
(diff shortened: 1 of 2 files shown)


client/runtime/src/test/java/io/quarkiverse/openapi/generator/providers/UrlPatternMatcherTest.java
```diff
@@ -32,7 +39,16 @@ private static Stream<Arguments> providePathsThatMatch() {
...
+                Arguments.of("/pets/{pet-id}/types/{type-id}", "/pets/1/types/2"),
+                Arguments.of("/repos/{ref}", "/repos/prefixed/cool.sw"),
+                Arguments.of("/repos/{ref}/pepe", "/repos/prefixed/cool.sw/pepe"),
+                Arguments.of("pepe/pepa/pepu", "pepe/pepa/pepu"));
+    }
+
+    private static Stream<Arguments> providePathsThatNotMatch() {
+        return Stream.of(
+                Arguments.of("/pets/{id}", "/pes/1"),
+                Arguments.of("/{id}/pepe", "/1/2/pep"));
```
gmunozfe, line 51: In this case it doesn't match because "pep" vs "pepe", but however
`Arguments.of("/{id}/pepe", "/1/2/pepe"));`
matches, is it correct? The placeholder should only contain one id, not two, wdyt?

PR #1034, opened 2025-03-06
WIP: feat: support RestResponse generation for reactive clients

### LINKED ISSUE #1033: Support org.jboss.resteasy.reactive.RestResponse<>
When using the reactive client it is possible to return a RestResponse. The RestResponse enables you to both get information about the response and keep the type safety of using a model. That's why I'd like this plugin to be able to use the RestResponse. This feature has already been requested here: #825 but the issue became stale.

pickbreaker:
Closes #1033


pickbreaker:
I am currently a little stuck on how to correctly implement the tests such that the RestResponse APIs are only generated if the maven profile ist reasteasy-reactive, because otherwise they throw a type error because of the missing resteasy-reactive dependency.

hbelmiro:
> I am currently a little stuck on how to correctly implement the tests such that the RestResponse APIs are only generated if the maven profile ist reasteasy-reactive, because otherwise they throw a type error because of the missing resteasy-reactive dependency.

@pickbreaker I'm not following.
Are you trying to return `org.jboss.resteasy.reactive.RestResponse` when using the classic profile?

pickbreaker:
No. In the integration tests the same code is generated no matter the profile. But this config option should only be supported on reactive clients like with the mutiny option. Now when the reactive code is generated when running the tests in classic profile the compilation of course fails. I tried only generating the api for when the profile is reactive, but could not find a clean way to do so.

PR #1104, opened 2025-04-10
Fix #1088 - Add support for custom credentials provider in runtime

ricardozanini:
Many thanks for submitting your Pull Request :heart:!

Fix #1088.

Please make sure that your PR meets the following requirements:

- [ ] You have read the [contributors guide](CONTRIBUTING.md)
- [ ] Your code is properly formatted according to [our code style](CONTRIBUTING.md#code-style)
- [ ] Pull Request title contains the target branch if not targeting main: `[0.9.x] Subject`
- [ ] Pull Request contains link to the issue
- [ ] Pull Request contains link to any dependent or related Pull Request
- [ ] Pull Request contains description of the issue
- [ ] Pull Request does not include fixes for issues other than the main ticket

<details>
<summary>
How to backport a pull request to a different branch?
</summary>

In order to automatically create a **backporting pull request** please add one or more labels having the following format `backport-<branch-name>`, where `<branch-name>` is the name of the branch where the pull request must be backported to (e.g., `backport-quarkus2` to backport the original PR to the `quarkus2` branch).

> **NOTE**: **backporting** is an action aiming to move a change (usually a commit) from a branch (usually the main one) to another one, which is generally referring to a still maintained release branch. Keeping it simple: it is about to move a specific change or a set of them from one branch to another.

Once the original pull request is successfully merged, the automated action will create one backporting pull request per each label (with the previous format) that has been added.

If something goes wrong, the author will be notified and at this point a manual backporting is needed.

> **NOTE**: this automated backporting is triggered whenever a pull request on `main` branch is labeled or closed, but both conditions must be satisfied to get the new PR created.
</details>

PR #1126, opened 2025-04-30
Use CredentialProvider in OAuth2 provider

gabriel-farache:

- [ ] You have read the [contributors guide](CONTRIBUTING.md)
- [ ] Your code is properly formatted according to [our code style](CONTRIBUTING.md#code-style)
- [ ] Pull Request title contains the target branch if not targeting main: `[0.9.x] Subject`
- [ ] Pull Request contains description of the issue
- [ ] Pull Request does not include fixes for issues other than the main ticket

This PR 

1. introduces tests for the Oauth2 provider (both Classic and Reactive depending on the delegate implementation used)
2. make use of the newly introduced `CredentialProvider` interface instance when setting the token in the header
  -  the behaviour is not a `get` as for the other providers but a `set` as the access token is first generated by the delegate and then set in the header. The reactive delegate and its Mutiny (async) usage force us to have all in one (getting the token and setting the header) to avoid losing the benefit of releasing the thread while waiting for the token.  If we are fine with losing it, a `getToken` method can instead be defined in the `OidcClientRequestFilterDelegate` interface and then use it in a  `CredentialProvider`  `get` method for oauth2 instead of calling the `filter` from the delegate
 3. add a new plugin to sort the import (this is completely optional but I was  annoyed with the import order error)


client/runtime/src/main/java/io/quarkiverse/openapi/generator/providers/ConfigCredentialsProvider.java
```diff
@@ -63,4 +68,22 @@ public String getBearerToken(ClientRequestContext requestContext, String openApi
...
+    @Override
+    public void setOauth2BearerToken(ClientRequestContext requestContext, String openApiSpecId, String authName,
+            ThrowingConsumer<ClientRequestContext, IOException> filter) throws IOException {
+        if (AbstractAuthProvider.isTokenPropagation(openApiSpecId, authName)) {
+            String bearerToken = AbstractAuthProvider.getTokenForPropagation(requestContext.getHeaders(), openApiSpecId,
+                    authName);
+            requestContext.getHeaders().add(HttpHeaders.AUTHORIZATION,
+                    OidcConstants.BEARER_SCHEME + " " + AbstractAuthProvider.sanitizeBearerToken(bearerToken));
+        } else {
+            filter.accept(requestContext);
```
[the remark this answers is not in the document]
gabriel-farache: If I do that, it voids the usefulness of using the CredentialProvider: the `filter` method of the delegate is setting the token in the header, hence if the user wants to do it differently, it will not be possible.

If this method is called after the `filter` is called, then yes the user could actually override the previously set header. 
If we want to do that, the `set` would take as input arg the `accessToken` and will do
```
requestContext.getHeaders().add("Authorization", "Bearer " + accessToken);
``` 
I was not "fan" of such solution as it means some things will always be done even if they may not be useful. The default method will simply do nothing.

I'll give a try in a separate commit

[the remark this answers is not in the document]
gabriel-farache: done.

client/runtime/src/main/java/io/quarkiverse/openapi/generator/providers/AbstractAuthProvider.java
```diff
@@ -60,22 +60,33 @@ public String getName() {
...
-                .orElse(false);
+        return isTokenPropagation(getOpenApiSpecId(), getName());
     }
 
-    public String getTokenForPropagation(MultivaluedMap<String, Object> httpHeaders) {
-        String headerName = getHeaderName() != null ? getHeaderName() : HttpHeaders.AUTHORIZATION;
-        String propagatedHeaderName = propagationHeaderName(getOpenApiSpecId(), getName(), headerName);
+    public static String getTokenForPropagation(MultivaluedMap<String, Object> httpHeaders, String openApiSpecId,
+            String authName) {
+        String headerName = getHeaderName(openApiSpecId, authName) != null ? getHeaderName(openApiSpecId, authName)
```
ricardozanini, line 68: Why not just `getName()` instead of receiving the `authName`? Something doesn't seem right. Maybe we should remove this method from here and move it to a utility class instead.

client/runtime/src/main/java/io/quarkiverse/openapi/generator/providers/AbstractAuthProvider.java
```diff
@@ -88,4 +99,15 @@ public final String getCanonicalAuthConfigPropertyName(String authPropertyName)
     public static String getCanonicalAuthConfigPropertyName(String authPropertyName, String openApiSpecId, String authName) {
         return String.format(CANONICAL_AUTH_CONFIG_PROPERTY_NAME, openApiSpecId, authName, authPropertyName);
     }
+
+    public static boolean isTokenPropagation(String openApiSpecId, String authName) {
```
ricardozanini, line 103: Same here regarding the `authName`. These methods now seems loosely coupled to this class, meaning that they might not be here.

client/runtime/src/main/java/io/quarkiverse/openapi/generator/providers/ConfigCredentialsProvider.java
```diff
@@ -63,4 +63,10 @@ public String getBearerToken(ClientRequestContext requestContext, String openApi
                 .orElse("");
     }
 
+    @Override
+    public void setOauth2BearerToken(ClientRequestContext requestContext, String accessToken, String openApiSpecId,
```
ricardozanini, line 67: In this case, you can create another interface that is a child of `CredentialsProvider` and add this method only to that interface.

client/runtime/src/main/java/io/quarkiverse/openapi/generator/providers/AbstractAuthProvider.java
gabriel-farache: that's because the `authName` is not a static field and we are in a static method

> Maybe we should remove this method from here and move it to a utility class instead.

Yes as this is used in different places

client/runtime/src/main/java/io/quarkiverse/openapi/generator/providers/ConfigCredentialsProvider.java
gabriel-farache: I just had an epiphany: if we are going to get the token previously set from the headers, then we can just use a `get` method.
After the `filter` is called, I call the `get` method from the credential provider and I store the token in a variable, as it is done in the others providers
Then at the end of the method, I set the header with this variable. In case token propagation is enabled, there is no way to override it (as for the other providers) but otherwise users can override what was set by the delegate.

And the default `get` method is simply retrieving the token set in the authorization header.

I did that using gh codespace and the format command is failing, hence the failing CI, if you like this approach I will fix it tomorrow (sorry for the live brainstorming)

client/runtime/src/main/java/io/quarkiverse/openapi/generator/providers/AbstractAuthProvider.java
gabriel-farache: same reason as above: those are instance field, not class field/variable so I cannot use them in a static context

ricardozanini: Oh misread the static

client/oidc/src/main/java/io/quarkiverse/openapi/generator/oidc/providers/OAuth2AuthenticationProvider.java
```diff
@@ -31,12 +31,19 @@ public OAuth2AuthenticationProvider(String name,
...
-        if (isTokenPropagation()) {
-            String bearerToken = getTokenForPropagation(requestContext.getHeaders());
-            bearerToken = sanitizeBearerToken(bearerToken);
-            requestContext.getHeaders().add(HttpHeaders.AUTHORIZATION, OidcConstants.BEARER_SCHEME + " " + bearerToken);
+        String bearerToken;
+
+        if (this.isTokenPropagation()) {
+            bearerToken = this.getTokenForPropagation(requestContext.getHeaders());
         } else {
             delegate.filter(requestContext);
```
ricardozanini, line 39: Why is filtering happening only on this branch? I'm concerned why our tests didn't catch this. :(

client/runtime/src/main/java/io/quarkiverse/openapi/generator/providers/ConfigCredentialsProvider.java
```diff
@@ -63,4 +64,8 @@ public String getBearerToken(ClientRequestContext requestContext, String openApi
                 .orElse("");
     }
 
+    @Override
+    public String getOauth2BearerToken(ClientRequestContext requestContext, String openApiSpecId, String authName) {
```
ricardozanini, line 68: You're only using the `requestContext` parameter here.

client/oidc/src/main/java/io/quarkiverse/openapi/generator/oidc/providers/OAuth2AuthenticationProvider.java
gabriel-farache: what do you mean?

client/runtime/src/main/java/io/quarkiverse/openapi/generator/providers/ConfigCredentialsProvider.java
gabriel-farache: Yes but I wanted to keep the same signature as the other methods declared in the interface, in case the user, for they own implementation, need to access specific properties.
I can completely remove it and users can open PR to add new arguments if needed

ricardozanini: Yes, I think that's more doable. We can overload the interface if needed in the future.

client/oidc/src/main/java/io/quarkiverse/openapi/generator/oidc/providers/OAuth2AuthenticationProvider.java
ricardozanini: Yes.

ricardozanini: Indeed, we are stopping the filtering chain there. Yes, you can leave it as it is now. I'll investigate it in the future.

gabriel-farache:
@ricardozanini A new version will have to be released, even if this PR is not accepted in the end, to have your #1104 PR made available, do you know when this is planned?

ricardozanini (commented):
LGTM, can you please add an integration test validating this feature so we can proceed with the merge?

A new version is likely to happen as soon as we merge this PR. We don't have a restricted timeline. We release once we have a handful of contributions or any critical bugs fixed.

ricardozanini:
> @ricardozanini I added IT but they fail: it looks like the custom credential provider I added is not taken into account, do you have an idea why?

No :(

gabriel-farache:
@ricardozanini I had to use Quarkus test profile to load the custom credentials provider for the IT, now it works, PTAL again :)

client/runtime/src/main/java/io/quarkiverse/openapi/generator/providers/ConfigCredentialsProvider.java
gabriel-farache: @ricardozanini Now that I am implementing the oauth2 token exchange pattern I see that I need those `openApiSpecId` and `authName` to only perform the oauth2 token exchange if the user enabled it for the given spec and security scheme; I do not want to exchange the token for all OAUTH2 security scheme

Now, if I extend the interface to add this new method `String getOauth2BearerToken(ClientRequestContext requestContext, String openApiSpecId, String authName)` and that my custom class extends `ConfigCredentialsProvider` and implements the newly created interface, the issue lies in the invocation of this new method as the interface implementation is called by the providers and they call `String getOauth2BearerToken(ClientRequestContext requestContext)`

How do you want to proceed?

client/integration-tests/auth-provider/src/main/resources/application.properties
```diff
@@ -0,0 +1,48 @@
+# Authentication properties
+org.acme.openapi.security.auth.basic_auth/username=john
+org.acme.openapi.security.auth.basic_auth/password=test
+customAuth.additionalApiTypeAnnotations=@org.eclipse.microprofile.rest.client.annotation.RegisterClientHeaders;@org.eclipse.microprofile.rest.client.annotation.RegisterProvider(io.quarkiverse.openapi.generator.it.security.auth.DummyApiKeyAuthenticationProvider.class)
+
+# KOGITO-6458 (https://issues.redhat.com/browse/KOGITO-6458) - generate auth bindings even if security definition is missing
```
ricardozanini, line 6: This comment can be removed.

ricardozanini (commented):
I'll take another look later.

ricardozanini (approved):
LGTM

gabriel-farache:
@ricardozanini I added a data structure which embed all parameters that must be made available to the methods of the CredentialProvider.

I added lombok dependency to have the getter and builder generated, if that pose a problem, I will remove, let me know

ricardozanini:
Hey @gabriel-farache many thanks! Yes, please remove lombok.

gabriel-farache:
> Hey @gabriel-farache many thanks! Yes, please remove lombok.

done

client/runtime/src/main/java/io/quarkiverse/openapi/generator/providers/CredentialsProvider.java
```diff
@@ -12,36 +12,96 @@ public interface CredentialsProvider {
...
+
+    /**
+     * Gets the OAuth2 Bearer Token given the OpenAPI definition and security schema
+     *
+     * @param input the input data available to the method
+     * @return the Bearer Token to use when filtering the request
+     */
+    String getOauth2BearerToken(DataInput input);
+
+    class DataInput {
```
ricardozanini, line 52:
```suggestion
    class CredentialsContext {
```

Can you change it to a more specific name? WDYF about `CredentialsContext`?

gabriel-farache: Ah, yes, thx, I struggled with thinking of a good naming here, this one is better than mine!

ricardozanini (approved):

ricardozanini:
I'll ask for more eyes.

client/oidc/src/test/java/io/quarkiverse/openapi/generator/oidc/OAuth2AuthenticationProviderTest.java
```diff
@@ -0,0 +1,124 @@
...
+        try {
+            Mockito.lenient().doCallRealMethod().when(classicDelegate).filter(requestContext);
+        } catch (IOException e) {
+            throw new RuntimeException(e);
+        }
+        provider = createClassicProvider();
+
+    }
+
+    protected OAuth2AuthenticationProvider createClassicProvider() {
```
hbelmiro, line 72:
```suggestion
    private OAuth2AuthenticationProvider createClassicProvider() {
```

client/oidc/src/test/java/io/quarkiverse/openapi/generator/oidc/OAuth2AuthenticationProviderTest.java
```diff
@@ -0,0 +1,124 @@
...
+        return new OAuth2AuthenticationProvider(AUTH_SCHEME_NAME, OPEN_API_FILE_SPEC_ID, classicDelegate, List.of());
+    }
+
+    protected void assertHeader(MultivaluedMap<String, Object> headers, String headerName, String value) {
```
hbelmiro, line 76:
```suggestion
    private void assertHeader(MultivaluedMap<String, Object> headers, String headerName, String value) {
```

client/integration-tests/auth-provider/src/test/java/io/quarkiverse/openapi/generator/it/auth/KeycloakServiceMock.java
```diff
@@ -0,0 +1,79 @@
...
+        return "{\n" +
+                "    \"access_token\": \"" + KEYCLOAK_ACCESS_TOKEN + "\",\n" +
+                "    \"expires_in\": 300,\n" +
+                "    \"refresh_expires_in\": 1800,\n" +
+                "    \"refresh_token\": \"" + KEYCLOAK_REFRESH_TOKEN + "\",\n" +
+                "    \"token_type\": \"bearer\",\n" +
+                "    \"not-before-policy\": 0,\n" +
+                "    \"session_state\": \"" + KEYCLOAK_SESSION_STATE + "\",\n" +
+                "    \"scope\": \"email profile\"\n" +
+                "}";
```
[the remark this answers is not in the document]
gabriel-farache: Nice, I missed this new java 15 feature!
As I copied this from another test, I think I will apply the change to the source of my copy/paste too

client/integration-tests/auth-provider/src/test/java/io/quarkiverse/openapi/generator/it/auth/TokenExternalServicesMock.java
```diff
@@ -0,0 +1,72 @@
...
+
+import com.github.tomakehurst.wiremock.WireMockServer;
+
+import io.quarkus.test.common.QuarkusTestResourceLifecycleManager;
+
+public class TokenExternalServicesMock implements QuarkusTestResourceLifecycleManager {
+
+    public static final String AUTHORIZATION_TOKEN = "AUTHORIZATION_TOKEN";
+    public static final String SERVICE3_AUTHORIZATION_TOKEN = "BEARER_TOKEN";
+    public static final String token_external_service_MOCK_URL = "propagation-external-service-mock.url";
```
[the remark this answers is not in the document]
gabriel-farache: ah, thx, search and replace gone wrong here

hbelmiro (approved):
Thank you @gabriel-farache.

client/oidc/src/test/java/io/quarkiverse/openapi/generator/oidc/ReactiveOAuth2AuthenticationProviderTest.java
```diff
@@ -0,0 +1,124 @@
...
+    }
+
+    @Test
+    void filterReactive() throws IOException {
+        filter(provider, "Bearer " + ACCESS_TOKEN);
+    }
+
+    private void filter(OAuth2AuthenticationProvider provider, String expectedAuthorizationHeader) throws IOException {
+        provider.filter(reactiveRequestContext);
+        assertHeader(headers, HttpHeaders.AUTHORIZATION, expectedAuthorizationHeader);
```
gmunozfe, line 96: `filter` is a non-blocking method and then, immediately, you are checking the headers in the `assertHeader`.
Potentially header couldn't be there yet, making the test to be flaky. Of course it's needed by Quarkus but it's not testing-friendly.

In this case, you could have a `filterAsync` method in the `ReactiveOidcClientRequestFilterDelegate` provider, returning a `Uni<Void>`
And then use here 
`provider.filterAsync(reactiveRequestContext).await().indefinitely();` to make more reliable this test.

Wdyt?

client/integration-tests/auth-provider/src/test/java/io/quarkiverse/openapi/generator/it/auth/TokenWithCustomCredentialProviderTest.java
```diff
@@ -0,0 +1,40 @@
...
+@Tag("resteasy-classic")
+@TestProfile(TokenTestProfile.class)
+class TokenWithCustomCredentialProviderTest {
+
+    @ParameterizedTest
+    @ValueSource(strings = { "service1", "service2", "service3", "service5" })
+    void testService(String service) {
+        Map<String, String> headers = new HashMap<>();
+        // service token-external-service1 and token-external-service2 will receive the AUTHORIZATION_TOKEN
+        headers.put(HttpHeaders.AUTHORIZATION, AUTHORIZATION_TOKEN);
```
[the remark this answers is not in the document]
gabriel-farache: right, I removed the other entries and I missed this refactor, thx!

client/oidc/src/test/java/io/quarkiverse/openapi/generator/oidc/ReactiveOAuth2AuthenticationProviderTest.java
gabriel-farache: not sure how I could use a `CountDownLatch` as I do not want to add code just for testing purposes in the `filter` part handling the response
Plus, the `getTokens` method is mocked and should return instantly.

But I do agree, in rare case, there may be sync issue and the test may fail; to avoid that, I mocked the `RestClientRequestContext` that executes the `suspend` and `resume` methods which are used in the  `filter` . THen I added a simple `while-sleep` loop to wait until the request is resumed. The loop may not be the most efficient solution, but for this case I believe it's enough as the impact will be very low, WDYT?

ricardozanini:
@gmunozfe mind taking a look?

gabriel-farache:
@ricardozanini should this part of the doc https://github.com/quarkiverse/quarkus-openapi-generator/blob/main/docs/modules/ROOT/pages/includes/custom-auth-provider.adoc?plain=1#L30 be updated? In our IT the priority is higher than the default one and it appears that our custom classes are used instead of the default one

ricardozanini:
> @ricardozanini should this part of the doc https://github.com/quarkiverse/quarkus-openapi-generator/blob/main/docs/modules/ROOT/pages/includes/custom-auth-provider.adoc?plain=1#L30 be updated? In our IT the priority is higher than the default one and it appears that our custom classes are used instead of the default one

Yes, the docs are wrong. I forgot to update on my PR. Can you update it here?

client/oidc/src/main/java/io/quarkiverse/openapi/generator/oidc/providers/OAuth2AuthenticationProvider.java
```diff
@@ -30,14 +31,30 @@ public OAuth2AuthenticationProvider(String name,
         validateConfig();
     }
 
+    public OAuth2AuthenticationProvider(String name,
+            String openApiSpecId, OidcClientRequestFilterDelegate delegate, List<OperationAuthInfo> operations) {
+        this(name, openApiSpecId, delegate, operations, new ConfigCredentialsProvider());
```
ricardozanini, line 36: Sorry I merged without noticing this change. This is wrong, you need to get the credProvider from the CDI, or your custom code won't be injected.

client/runtime/src/main/java/io/quarkiverse/openapi/generator/providers/CredentialsProvider.java
```diff
@@ -12,36 +12,96 @@ public interface CredentialsProvider {
...
+
+    /**
+     * Gets the OAuth2 Bearer Token given the OpenAPI definition and security schema
+     *
+     * @param input the input data available to the method
+     * @return the Bearer Token to use when filtering the request
+     */
+    String getOauth2BearerToken(CredentialsContext input);
+
+    class CredentialsContext {
```
ricardozanini, line 52: Please make this its own class.

ricardozanini (commented):
@gabriel-farache can you please follow up a new PR with my late considerations? I'm sorry, I've merged too soon.

client/oidc/src/main/java/io/quarkiverse/openapi/generator/oidc/providers/OAuth2AuthenticationProvider.java
gabriel-farache: too late :) https://github.com/quarkiverse/quarkus-openapi-generator/pull/1176

Issue #1227, opened 2025-07-03
OIDC Client: Is it possible?


melloware:
@ricardozanini @hbelmiro apologize for opening this ticket but working with @mcruzdev i tried everything to figure this out and now i am starting to not think its possible.

1) I have a Quarkus REST server using OIDC authentication with KeyCloak. here is the server config.

```properties
quarkus.oidc.enabled=true
quarkus.oidc.client-id=my-client-id
quarkus.oidc.auth-server-url=http://localhost:8020/realms/myrealm
```

2. It generates this in the OpenAPI JSON:

```json
    "securitySchemes" : {
      "SecurityScheme" : {
        "type" : "openIdConnect",
        "openIdConnectUrl" : "http://localhost:8020/realms/myrealm/.well-known/openid-configuration",
        "description" : "Authentication"
      }
    }
```

3. Now I want to use OpenAPIGenerator to generate my client code.  I have read all the instructions: https://docs.quarkiverse.io/quarkus-openapi-generator/dev/client.html#_oauth2_authentication  I have tried everything and no matter what it errors with `cannot find package com..xxx.auth` as its not generating the .auth filter.

# Questions

Is OIDC supported?

If OIDC is not supported is the idea I need to turn off security `quarkus.openapi-generator.codegen.enable-security-generation=false` and create my OIDC client code and wire it into my generated code?

Feel free to close this ticket if its not valid but I think this will come up more and more as OIDC is popular.

ricardozanini:
@melloware, this is supposed to work. If not, it might be a side effect of changes in the latest versions. Since we decided to go full Quarkus and generate Synthetic Beans instead of predefined template files for auth, we have been experiencing flaws.

I don't have the bandwidth to implement this fix in the short term. @gabriel-farache, could you mind taking a look too, since you're working on security-related features? It might impact your use cases.

melloware:
I can provide a reproducer Quarkus server app if needed.

indiealexh:
Ok! After a lot of messy flailing around and clean up... Here it is! #1284

PR #1284, opened 2025-08-22
Add Security Scheme type: openIdConnect

indiealexh:
Fixes: https://github.com/quarkiverse/quarkus-openapi-generator/issues/1227

Add to the template and codegen to look for and use the openIdConnectMethods
Add Integration tests to verify support


indiealexh:
> @indiealexh do you need a patch release for this fix?

It would be nice, so I can clean up my api specs on oauth2 configs :D

PR #1381, opened 2025-11-18
feat: support RestResponse

### LINKED ISSUE #1033: Support org.jboss.resteasy.reactive.RestResponse<>

BB223:
Closes #1033

### DIFF
--- a/client/deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/CommonItemConfig.java
+++ b/client/deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/CommonItemConfig.java
@@ -68,7 +68,7 @@ public interface CommonItemConfig {
 
     /**
      * Defines if the methods should return {@link jakarta.ws.rs.core.Response},
-     * {@link org.jboss.resteasy.reactive.RestResponse} or a model. Default is {@code false}.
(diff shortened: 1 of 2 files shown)


client/deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/CommonItemConfig.java
```diff
@@ -67,10 +67,11 @@ public interface CommonItemConfig {
     Optional<String> additionalRequestArgs();
 
     /**
-     * Defines if the methods should return {@link jakarta.ws.rs.core.Response} or a model. Default is {@code false}.
+     * Defines if the methods should return {@link jakarta.ws.rs.core.Response},
+     * {@link org.jboss.resteasy.reactive.RestResponse} or a model. Default is {@code false}.
```
ricardozanini: What's the default now? `false`? What does this mean? We should clarify this a little bit better. 

> By default, it returns the model in the specification

Or something like that.

ricardozanini (commented):
Looks good, many thanks! Just a comment on the configuration you changed.

PR #1514, opened 2026-03-31
Skip generation if spec is unchanged

### LINKED ISSUE #878: Class generation from spec only when spec changed
When I start my quarkus application with mvn quarkus:dev the openapi-generator from quarkus always generates the proxies from the swagger file, no matter if the spec changed or not. This triggers the compile and increases the startup time of the dev session.
It would be nice when the generator has a change detection which only generates the proxies in case of a changed swagger file.
Thanks for any comments on this

CodeSimcoe:
Proposition to address #878

Skip generation if :
- property is enabled (maybe it could be enabled by default ? I've put it to `false` for now, for compatibility reasons)
- spec hasn't changed
- config hasn't changed


ricardozanini (commented):
Looks good! Many thanks for taking your time to help us :)

mcruzdev (changes requested):
Thanks a lot for this pull request @CodeSimcoe 👍🏼 

I left a few comments, see if makes sense 😃

mcruzdev (approved):

CodeSimcoe:
> Hey @CodeSimcoe thanks for the new updates, could you squash and to solve the conflicts?

Should be good

ricardozanini (approved):

PR #1705, opened 2026-06-16
Fix CVE-2026-42333: Prevent credential leakage via path-template-based authentication matching

ricardozanini:
## Vulnerability

  URL regex pattern matching incorrectly matched literal sibling paths to parameterized templates, causing credential leakage to unintended endpoints.

  **Example:**
  - Protected: `GET /repos/{ref}` (requires bearer token)
  - Unprotected: `GET /repos/health` (no security)
  - **Bug:** Regex `/repos/([^/]*)` matched BOTH paths, leaking credentials to `/repos/health`

  ## Fix

  ### 1. New `OperationIdFilter`
  Extracts operation metadata from `@OperationMarker` annotations and sets request context properties (`operation.path`, `operation.method`)
  - Uses standard MicroProfile `invokedMethod` property + reflection fallbacks
  - Compatible with both RESTEasy Classic and Reactive
  - Registered via `@RegisterProvider` in generated REST client interfaces

  ### 2. Updated `BaseCompositeAuthenticationProvider.canFilter()`
  Replaces URL regex matching with exact path-template matching
  - Matches by path template (e.g., `/repos/{ref}`) + HTTP method from OpenAPI spec
  - Fail-secure: if properties not set, authentication is not applied
  - `/repos/{ref}` ≠ `/repos/health` (exact string match prevents leakage)

  ### 3. Code generation
  Template updated to register `OperationIdFilter` when security is enabled

  ## Testing

  - ✅ All 100 runtime unit tests pass
  - ✅ Integration tests pass with RESTEasy Classic
  - ✅ Integration tests pass with RESTEasy Reactive
  - ✅ New `SecurityVulnerabilityCVE20264233Test` validates the fix

  ## Security

  Path-template matching eliminates regex ambiguity and prevents credentials from being sent to unintended endpoints.

### DIFF
--- a/client/deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/OperationIdFilterRegistrationProcessor.java
+++ b/client/deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/OperationIdFilterRegistrationProcessor.java
@@ -1,0 +1,26 @@
+package io.quarkiverse.openapi.generator.deployment;
+
+import io.quarkiverse.openapi.generator.providers.OperationIdFilter;
+import io.quarkus.arc.deployment.AdditionalBeanBuildItem;
+import io.quarkus.deployment.annotations.BuildStep;
+import io.quarkus.deployment.builditem.FeatureBuildItem;
+
+/**
+ * Quarkus build step to ensure OperationIdFilter is available as a CDI bean.
+ * This is needed for RESTEasy Reactive which may not honor @RegisterProvider annotations
+ * on provider classes referenced in REST client interfaces.
+ */
+public class OperationIdFilterRegistrationProcessor {
+
+    private static final String FEATURE = "openapi-generator-operation-id-filter";
+
+    @BuildStep
+    FeatureBuildItem feature() {
+        return new FeatureBuildItem(FEATURE);
+    }
+
+    @BuildStep
+    AdditionalBeanBuildItem registerOperationIdFilter() {
+        return AdditionalBeanBuildItem.unremovableOf(OperationIdFilter.class);
+    }
+}
--- a/client/deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/OperationIdFilterRegistrationProcessor.java
+++ b/client/deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/OperationIdFilterRegistrationProcessor.java
@@ -3,7 +3,6 @@ package io.quarkiverse.openapi.generator.deployment;
 import io.quarkiverse.openapi.generator.providers.OperationIdFilter;
 import io.quarkus.arc.deployment.AdditionalBeanBuildItem;
 import io.quarkus.deployment.annotations.BuildStep;
-import io.quarkus.deployment.builditem.FeatureBuildItem;
 
 /**
  * Quarkus build step to ensure OperationIdFilter is available as a CDI bean.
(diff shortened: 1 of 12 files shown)


ricardozanini:
@anir0y mind taking a look since you reported it?

anir0y:
Thanks for the quick turnaround, @ricardozanini. I reviewed the PR and the approach correctly resolves what I reported.

The original issue was that `canFilter` re-derived the operation from the request URL with a regex `anyMatch` and no exact-operation precedence, so a protected template like `GET /repos/{ref}` (`^/repos/([^/]*)$`) also matched the unprotected literal sibling `GET /repos/health`, attaching the credential to a spec-unauthenticated endpoint.

This PR fixes the root cause rather than the symptom:

- `OperationIdFilter` resolves the actual invoked operation from its `@OperationMarker` and stores the template path + method on the request context, and runs at `@Priority(AUTHENTICATION - 1000)` so the data is set before the auth provider reads it. Ordering is correct.
- `canFilter` now matches on the exact OpenAPI template path + method (`o.getPath().equals(requestPath) && o.getHttpMethod().equals(requestMethod)`), not a regex against the resolved URL. `"/repos/{ref}"` is no longer equal to `"/repos/health"`, so the sibling collision is gone. Keying on the template rather than the resolved path is the right choice.
- The fail-secure default (no credentials applied when the operation context is absent) is the correct call: an unexpected gap denies auth instead of over-attaching it.
- Per-spec scoping (`openApiSpecId` / `operationsToFilter()`) keeps the exact-equality match from colliding across specs.
- The regression test covers the exact reported pair (`/repos/{ref}` vs `/repos/health`).

LGTM on the security fix. One thing worth confirming before merge, more functional than security:

- The fail-secure branch means any invoked REST-client method that reaches `canFilter` without an `@OperationMarker` (a hand-written sub-resource method, or any operation the generator does not annotate) will now go out with no authentication, silently. Could you confirm the generator emits `@OperationMarker` on every generated operation, and note the expected behavior for hand-authored client methods? That keeps this from becoming a quiet auth regression for those calls.

Minor: the test class is named `SecurityVulnerabilityCVE20264233Test`, but the CVE is `CVE-2026-42333` (looks like a digit was dropped). Worth aligning the name for discoverability.

Thanks for crediting the report and for the thorough fix.

ricardozanini:
> Could you confirm the generator emits @OperationMarker on every generated operation, and note the expected behavior for hand-authored client methods? That keeps this from becoming a quiet auth regression for those calls.

Yes, the generator adds the `@OperationMarker ` annotation to every generated resource if the user option is to use our auth filters.

For hand-authored client methods, that's out of our scope since we can't control these and we won't add authentication details to them.

ricardozanini:
@mcruzdev please take a look too

client/deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/OperationIdFilterRegistrationProcessor.java
```diff
@@ -0,0 +1,26 @@
...
+import io.quarkus.deployment.builditem.FeatureBuildItem;
+
+/**
+ * Quarkus build step to ensure OperationIdFilter is available as a CDI bean.
+ * This is needed for RESTEasy Reactive which may not honor @RegisterProvider annotations
+ * on provider classes referenced in REST client interfaces.
+ */
+public class OperationIdFilterRegistrationProcessor {
+
+    private static final String FEATURE = "openapi-generator-operation-id-filter";
```
mcruzdev: Why we need a new `FeatureBuildItem` being produced from client extension, once normally the feature is 1:1 to the extension?

mcruzdev (approved):
LGTM, I left a comment about the new `FeatureBuildItem`, I think it is not necessary.

client/deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/OperationIdFilterRegistrationProcessor.java
ricardozanini: Yeah, this doesn't make sense, it's a leftover. I'll remove it.

PR #1714, opened 2026-06-20
fix: correct wording in documentation for Apicurio usage

matheusandre1:
a small error in the documentation


mcruzdev (approved):
Good catch!

PR #1735, opened 2026-07-24
feat: support for generating a single API class with single-api-name option

### LINKED ISSUE #360: Feature request, generate single api class from swagger
Currently multiple api classes are generated, one api class per tag defined in swagger file.

Would be perfect if there is an option to define single class name where all the api methods will be generated, no matter what tags are defined in the file. I am migrating from our custom code generation plugin where this feature was supported and it would make the transition much easier. Adjusting the swagger files is not an option for us.

I am open to implement this enhancement with a bit of support from community.

Thanks for the feedback

matheusandre1:

Closes : #360 

By default, the generator creates one API class per tag defined in the OpenAPI specification. This PR adds a new per-spec configuration property, `single-api-name`, that lets users generate all operations into a single API class, regardless of the tags declared in the spec file:

```properties
quarkus.openapi-generator.codegen.spec.my_openapi_yaml.single-api-name=MyService
```

The value is applied as the tag for every operation, so the example above produces a single `MyServiceApi` class containing all operations (the `Api` suffix can still be customized with `api-name-suffix`). This is equivalent to the OpenAPI Normalizer rule `SET_TAGS_FOR_ALL_OPERATIONS`; if both are configured, `single-api-name` takes precedence.

- [X] You have read the [contributors guide](CONTRIBUTING.md)
- [X] Your code is properly formatted according to [our code style](CONTRIBUTING.md#code-style)
- [X] Pull Request title contains the target branch if not targeting main: `[0.9.x] Subject`
- [X] Pull Request contains link to the issue
- [X] Pull Request contains link to any dependent or related Pull Request
- [X] Pull Request contains description of the issue
- [X] Pull Request does not include fixes for issues other than the main ticket

### DIFF
--- a/client/deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/CodegenConfig.java
+++ b/client/deployment/src/main/java/io/quarkiverse/openapi/generator/deployment/CodegenConfig.java
@@ -72,6 +72,7 @@ public interface CodegenConfig extends GlobalCodegenConfig {
         IMPORT_MAPPINGS("import-mappings"),
         SCHEMA_MAPPINGS("schema-mappings"),
         NORMALIZER("open-api-normalizer"),
+        SINGLE_API_NAME("single-api-name"),
         RETURN_RESPONSE("return-response"),
         ENABLE_SECURITY_GENERATION("enable-security-generation"),
         CONFIG_KEY("config-key"),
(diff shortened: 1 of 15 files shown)


mcruzdev (changes requested):
Nice pull request, I would like to have the value of the property as the class name. 

If the user set `*.single-api-name=MyService` the generated class must be `MyService`, not `MyServiceApi` (of course, it should be documented). 

Additionally, we need to be sure that the value set by the user is a valid Java class name, could you create a test with this scenario?

mcruzdev (approved):
LGTM!

@ricardozanini Do you mind to take a look?

